PR #19, opened 2017-07-12
Fix Chrome crashes by using host shared memory

### LINKED ISSUE #12: Chrome behat runs crashing
As discovered by Eloy: 
https://pastebin.com/3NUAjbmn

I haven't really looked into chrome or even using a named tag, so needs some work.

stronk7:
Ref: https://github.com/SeleniumHQ/docker-selenium#running-the-images

(Should fix #12 - there are execution results there)


danpoltawski:
It seems symphony/github api is suffering on the travis runs at the moment and jobs failing. Think we'll need to try again when is stabilised.

Issue #10, opened 2017-07-05
Persistance, for 'manual' testing


danpoltawski:
For doing manual testing, we don't want our docker environments to be completely torn down and recreated each time.

Need to investigate this:

1. I think docker-compose does some sort of volume persistance itself, especially if you use `stop` rather than `down`. But I have not looked into this at all
2. We might want explict mounted data directories

So far i've been focused initially on quick manual tests and automated testing and haven't examined this case at all, but manual testing is very much part of the usecase for this project.

abias:
Hi Dan,

> For doing manual testing, we don't want our docker environments to be completely torn down and recreated each time.

I completely agree with this need for our use cases.

> I think docker-compose does some sort of volume persistance itself, especially if you use stop rather than down. But I have not looked into this at all

I'm not a Docker expert yet, but I have seen than with
`bin/moodle-docker-compose stop`
and
`bin/moodle-docker-compose start`
you can stop and start the docker containers and the containers will still have their data, even if the whole docker environment is shut down and restarted in the meantime.

So, do you think there is anything left to do or unsure?
If not, I would create a PR to add this information to the README.md.

> We might want explict mounted data directories

Could you please elaborate what you mean with this sentence so that I can understand the problem you are seing?

Thanks,
Alex

scara:
Hi @abias,
trying to anticipate Dan's reply, that means assigning real folders (i.e. somewhere out of the container, e.g. in the host) where to safely store what required to persist the tests setup (or some hacking done while investigating a failure) i.e. the database and the fixtures as deployed into Moodle i.e. the so called moodle data.
Read more at https://docs.docker.com/compose/compose-file/compose-file-v2/#volumes-volume_driver.

Since this setup is for dev/test purposes I'm keen to second your idea of PR w/o the need to expore more optional params to give persistance to the environment: anyone playing with `down` will be warned about the lacking of persistance by design.
Indeed, the only drawback I found w/ this lacking is when switching among different DBs which requires a `down`=>`export MOODLE_DOCKER_DB=<other_db_type>`=>`up -d` sequence but I'm fine with missing the _status_ while switching.

HTH,
Matteo

abias:
Hi Matteo,

thank you for your feedback.

[...]

PR #21, opened 2017-07-13
Add help for manual testing

xow:
I've been using this for manual testing as well as behat / unit testing. I went to set it all up on my new environment and completely blanked as to why I couldn't use the site manually.

Because we create a config.php the error message is confusing (cannot access database) and I couldn't remember that we need to do cli DB install. (Even though I guess I did it last time).

Feel free to reject this pull request if we don't want to encourage manual testing :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,10 @@ Started at 25-05-2017, 19:04
 15 steps (15 passed)
 1m35.32s (41.60Mb)
 
+# Install for manual testing (optional)
+bin/moodle-docker-compose exec webserver php admin/cli/install_database.php --agree-license --fullname="Docker moodle" --shortname="docker_moodle" --adminpass="test" --adminemail="admin@example.com"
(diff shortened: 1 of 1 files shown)


danpoltawski:
Thanks John, this is definitely to be encouraged. Although need to work out solution to #10 better.

Note if you visit in web browser it'll bring the installer properly though?

xow:
@danpoltawski It was interesting, because since the config.php is copied in, moodle just says "cannot access database" when I access it via the browser. (on mac) I don't think I did anything beyond the normal instructions.

I don't remember having to do the db install when I tried on linux though. Perhaps something weird in my mac environment?

danpoltawski:
@xow that doesn't sound correct. When I do that I get prompted for the installer (this is what our install.php does, creates config.php and then hands over to web installer). I wonder if it could be a recent master regression?

PR #23, opened 2017-07-20
Add behat faildump and expose it in a nice way

andrewnicols:
This change maps an optionally provided directory to a location on disk, and
ensures that the Moodle configuration is in place to use it for faildumps and
screenshots.  # Requesting a pull to danpoltawski:master from
andrewnicols:faildump_map


andrewnicols:
I can't see any relation between the change here and the fail experienced (A MySQL test failure).
The only change I can see in the image is MariaDB changing from version `10.2.6` to `10.2.7`.

danpoltawski:
Thanks for the patch Andrew, I would like us to go a bit further than this though:

- I think we should enable the faildump directory by default (there is nothing more annoying than starting a behat run to find it not enabled and not wanting to stop a run)
- I think we should expose it via the webserver maybe in an alias likes /fails/ with directory listing on, so you can just browse through your browser (this will probably require a change in the php image)
- I'm not sure about adding the variable, in general I am keen for us to avoid adding too many variables because it just introduces complexity. I understand there would be a need for a persistent directory in some situations, but perhaps it could be achieved with `docker cp` and we can avoid adding another var

Agree mariadb is unrelated to this issue

danpoltawski:
Have created #25 for the mariadb problem

danpoltawski:
I'm closing this in favour of  #31

PR #24, opened 2017-07-20
Stop default mapping of the web server port

andrewnicols:
If you want to run multiple jobs at once, you cannot specify a port mapping for the web server.

This could be achieved in several ways. The most obvious ways being:
* Do not map a port by default, but respect a port specified in an environment variable;
* Require the caller to write a YML file overriding the port configuration; or
* Have the calling code guess a free port.

I opted for the first option because neither PHPUnit, nor Behat _require_ that the site be installed and available, and the specification of an optional port number in an environment variable is a low barrier to adoption.

I did consider having a default value, but then something more intelligient would be required to unset it for parallel runs.


andrewnicols:
FYI, I'm waiting on the other issues here before proceeding here.
The principle is to be (part of) the solution for runnign multiple tests concurrently (the other part is to specify a project-name with `-p example` to docker-composer).

danpoltawski:
Need to consider manual testing with this (currently hardcodes `'http://localhost:8000'` into wwwroot).

Part of me likes the idea of a reverse proxy vm, which maps running containers on a single port (like mysql.docker.local), but I can't think of a good way to orchestrate this really. I wonder if docker-compose provides some tools we could use here.

andrewnicols:
I've considered it, in that it will append the port number if it is specified. Really it should use the entire hostname in the `config.php` too.

The current patch will allow for the dynamic port number that Docker determines.

Unfortunately, although we can use a dynamic port number in the yml file, we can't get the port number to add it to the environment. I suspect that the only way to make the dynamic port work would be to add a reverse proxy somewhere in the chain.

The issue with this is that we'd need to _additionally_ add all of the containers to the same network. Each set of images created with `docker-compose` creates a new bridge network for those VMs. The default bridge doesn't support the internal DNS, but it _can_ talk to all VMs (just not by name).

Another option would be to guess a port number from within your scripts, but that seems perilous and problematic - especially for Windows.

This patch doesn't prevent anything - you can still reach the VM using the dynamic port, e.g. `localhsot:32784`, but unfortunately it doesn't quite work as expected without adding in some form of reverse proxy or magic to the config.php (like the one I've written for ngrok -https://gist.github.com/andrewnicols/7c9d71fb21234b3479e4d9cae8d16074).

andrewnicols:
Note: It _currently_ isn't possible to get the details of a running container from within it without nasty hackery, but I understand that's a work-in-progress.

andrewnicols:
I've pushed a new patch which maintains the default behaviour, but allows the port to be unset.

danpoltawski:
Andrew and I had discussion about the mailcatcher thing and Andrew had the great idea of taking the mailcatcher port out of the equation by adding some apache config to proxy to mailcatcher - which I think is a great idea.

I'm really aggressively keen to try and avoid us adding too many options/variables to keep things simple (as I mentioned on another issue, if we don't keep it simple, people may as well edit their own compose.yml file)

andrewnicols:
Alas, this wasn't as easy as I had hoped because mailcatcher is very tied to running on a port and uses absolute URLs everywhere. In fact, I found it was getting harder than it was worth with the combination of content rewriting, and web socket tunnelling that I was unable to get it to work reliably.

Instead I have looked at alternatives to mailcatcher and switched to MailHog. It essentially does the same thing, but is written in Go and is more amenable to being reverse proxied.

I'm happy to move this part of the request to a new issue if you prefer.

bin/moodle-docker-compose.cmd
```diff
@@ -13,6 +13,7 @@ POPD
 SET COMPOSE_CONVERT_WINDOWS_PATHS=true
 
 SET DOCKERCOMPOSE=docker-compose -f "%BASEDIR%\base.yml"
+SET DOCKERCOMPOSE=%DOCKERCOMPOSE% -f "%BASEDIR%\service.mail.yml"
```
scara: Hi @andrewnicols,
is there any reason to keep line 15?

TIA,
Matteo

scara: Ignore my comment :blush:: you're properly use multiple configuration files via `-f`.

Apologies for the noise!
Matteo

base.yml
```diff
@@ -4,7 +4,7 @@ services:
     image: danpoltawski/moodle-php-apache:7.1
     depends_on:
       - db
-      - mail
+      - mailhog
```
danpoltawski: Can we put this into `services.mail.yml`? It's a bit unclean in my mind to depend on something which isn't included in the same file.

andrewnicols:
Good thinking. I wasn't entirely sure on the purpose of the `depends_on` but I've just read up on it.

danpoltawski:
Thanks Andrew, merged - this looks great. I've not yet merged to all the other branches because download server is down and is affecting the unit tests unfortunately..

danpoltawski:
Have merged to the other branches now

PR #30, opened 2017-07-26
Use mailcatcher for email testing

### LINKED ISSUE #28: Mailcatcher SMTP server
We should add mailcatcher as the default smtp server so emails can be browsed - affects #27

danpoltawski:
This could be an optional depdency, but I think the simplicity benefits
of having it run by default are worthwhile. (Can be changed in future to
be disabled, or replaced with a more eleborate smtp setup).

Fixes #28


scara:
Hi @danpoltawski,
you should add a new dependency too, in the `webserver`, here: https://github.com/danpoltawski/moodle-docker/blob/8730f6e19d3deff417f774433c6786d8d89c2839/base.yml#L7 .

HTH,
Matteo

danpoltawski:
Good point, thanks - have updated the patch

PR #31, opened 2017-08-02
Add behat faildump path

### LINKED ISSUE #22: Add behat faildump and expose it in a nice way
Add a behat faildump path and expose it (perhaps through web browser) and consider any other tricks to ensure you can see what's going on with in progress behat runs

danpoltawski:
(and expose it via the webserver)

Fixes #22


danpoltawski:
Note this change depends on the new moodle-php-apache images

PR #32, opened 2017-08-03
Add some debug options to default config

### LINKED ISSUE #27: Update config.php to enable dev options..
How did I miss this? We should enable all useful config debugging options - some ones which people don't know about in:
https://www.slideshare.net/poltawski/moodle-moot-usa-16-tools-and-tips-for-moodle-developers-mootus16

This could be controlled by an environment flag, but I think debug mode should be enabled by default - this is intended for devs primarily.

danpoltawski:
We might want to control this by a flag in future

Fixes #27

### DIFF
--- a/config.docker-template.php
+++ b/config.docker-template.php
@@ -23,6 +23,15 @@ $CFG->admin     = 'admin';
 $CFG->directorypermissions = 0777;
 $CFG->smtphosts = 'mailhog:1025';
 
+// Debug options - possible to be controlled by flag in future..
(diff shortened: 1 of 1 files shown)


config.docker-template.php
```diff
@@ -23,6 +23,15 @@
 $CFG->directorypermissions = 0777;
 $CFG->smtphosts = 'mailhog:1025';
 
+// Debug options - possible to be controlled by flag in future..
```
scara: Typo: trailing dot.

danpoltawski: Ha, it was an intentional something more...

Issue #33, opened 2017-08-16
Moodle mapping to localhost not always fine


scara:
If you run the container host engine on a machine different from your local dev environment, Moodle [mapping to `localhost`](https://github.com/danpoltawski/moodle-docker/blob/246e98a1b2cf83723304d42241b16e3bd4ff9440/config.docker-template.php#L16) is fine for automated tests "only".

I know that you don't like too many env variables :wink: but what about using `$_SERVER` (e.g. `SERVER_NAME` or `HTTP_HOST` or `SERVER_ADDR`) or kind-of-`gethostname()` - via script, to get the name of the host and not the one of the container, unroutable - here being the scope of this composition mainly for dev/test purposes i.e. no need to preserve the uniqueness of URLs eventually stored in different "sessions"?
I'd be happy to contribute here according to your thoughts.

Pinging @andrewnicols too, since his [proxy solution](https://github.com/danpoltawski/moodle-docker/blob/894365bb597f9ab6976f64d7d2f4a2fdd7ba9b2d/assets/web/apache2_mailhog.conf) for some ancillary services has unlocked the access of those services from `localhost` only.

TIA,
Matteo

danpoltawski:
I actually think this is decent case for an ENV variable (defaulting to localhost)

scara:
TNX, will do 😉 !

PR #37, opened 2017-09-08
Adding phpMyAdmin support for MySQL and MariaDB databases.

rlorenzo:
I found it useful to have phpMyAdmin during Moodle development. Hopefully others will as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,10 @@ Started at 25-05-2017, 19:04
 bin/moodle-docker-compose exec webserver php admin/cli/install_database.php --agree-license --fullname="Docker moodle" --shortname="docker_moodle" --adminpass="test" --adminemail="admin@example.com"
 # Access http://localhost:8000/ on your browser
 
+# Access phpMyAdmin (for MySQL and MariaDB databases)
+
+If you are running MySQL or MariaDB, you can navigate to http://localhost:8001 to access phpMyAdmin for database management.
+
 # Shut down containers
 bin/moodle-docker-compose down
 ```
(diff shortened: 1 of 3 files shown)


scara:
Just sharing a thought: should any additional web-based _service_ be proxied to avoid using new TCP ports? See https://github.com/moodlehq/moodle-docker/blob/master/assets/web/apache2_mailhog.conf and https://github.com/moodlehq/moodle-docker/pull/24#issuecomment-318362082.

HTH,
Matteo

danpoltawski:
> should any additional web-based service be proxied to avoid using new TCP ports?

Yes, I think this is a good idea. (i’m not completely sold on us adding phpmyadmin by default)

scara:
> I’m not completely sold on us adding phpmyadmin by default

I'm but I'd vote for https://www.adminer.org/ in this case, being cross-DBs (MySQL, PostgreSQL, SQLite, MS SQL, Oracle, SimpleDB, Elasticsearch, MongoDB).

It could be useful when your dev env is not hosted in the same machine where the docker daemon is running (mine :wink:, #33); e.g. when working w/ https://tracker.moodle.org/browse/MDL-59100 (and related) I've used the CLI client, from the `db` container, to connect to the DB and a Web-based solution would have speed up some investigation works.

HTH,
Matteo

danpoltawski:
Hi Rex,

We talked about this a bit in HQ and I'm going to close this pull request because I didn't really think it was the right choice and i couldn't get others to support the idea either.

There are a few things against it:

- It is just one method of accessing database which is not universally applicable (many will use other tools like GUI applications, or if you are like me, the CLI tools only)
- This only supports the one db engine rather than something like adminer which could apply to all the different db engines
- It seems like installing local_adminer in your codebase would not require this adding to the docker images

Overall, it's a sort of 'opinionated close' that I don't think it's worth adding this to the images - but I think it's better that we close this than keep it open in limbo.

PR #38, opened 2017-09-08
Adding phpMyAdmin support for MySQL and MariaDB databases.

rlorenzo:
For the php71 branch.

Same as https://github.com/moodlehq/moodle-docker/pull/37


danpoltawski:
I'm going to close these branch pull requests - I think it's better if we just work on the one pull request and then merge onto the branches.

(I should add a pull request template to this repo to explain this as I think it stops the pull requests getting overwhelmed)

PR #41, opened 2017-09-12
Adding in support for VNC to view running Behat tests.

rlorenzo:
We had some of our talented student developers tackle on enabling someone to view Behat tests using VNC. Check it out.

### DIFF
--- a/base.yml
+++ b/base.yml
@@ -20,6 +20,9 @@ services:
       POSTGRES_PASSWORD: "m@0dl3ing"
       POSTGRES_DB: moodle
   selenium:
-    image: selenium/standalone-firefox:2.53.1
+    image: selenium/standalone-firefox-debug:2.53.1
+    build: ./selenium/
+    ports:
+      - 5900:5900
     volumes:
       - "${MOODLE_DOCKER_WWWROOT}:/var/www/html:ro"
--- a/README.md
+++ b/README.md
@@ -45,6 +45,12 @@ Started at 25-05-2017, 19:04
 15 steps (15 passed)
 1m35.32s (41.60Mb)
 
+# Using VNC to view Behat tests
+
+1. Download a VNC client: https://www.realvnc.com/en/connect/download/viewer/
+2. With the containers running, enter 0.0.0.0:5900 as the port in VNC Viewer. You will be prompted for a password. The password is 'test' by default, but you can change it by editing ./selenium/Dockerfile
+3. You should be able to see an empty Desktop. When you run any Behat tests a browser will popup and you will see the tests execute.
+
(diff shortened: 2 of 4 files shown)


base.yml
```diff
@@ -20,6 +20,9 @@ services:
       POSTGRES_PASSWORD: "m@0dl3ing"
       POSTGRES_DB: moodle
   selenium:
-    image: selenium/standalone-firefox:2.53.1
+    image: selenium/standalone-firefox-debug:2.53.1
+    build: ./selenium/
```
danpoltawski: I think that image and build are conflicting here? If this is working I assume that build is winning as the last one.

danpoltawski:
Hi Rex/Student Developers,

1) My gut feel is that we don't want this by default, but should control it with some debug option. This might get complex.
2) I don't quite understand why you are doing the build and setting the vnc password to test - https://github.com/SeleniumHQ/docker-selenium#debugging suggests you can just use the debug images and the password is `secret`

abias:
While I haven't tested this patch myself, I would really be interested in this feature for manual testing. 

> My  gut feel is that we don't want this by default, but should control it with some debug option. 

Or perhaps you could control that with another environment variable for /bin/moodle-docker-compose

danpoltawski:
> While I haven't tested this patch myself, I would really be interested in this feature for manual testing.

Why not use your own browser for that? I don't feel like using selenium node is best solution for this.

> Or perhaps you could control that with another environment variable for /bin/moodle-docker-compose

Yes, sorry I wasn't clear - that is what I meant.

base.yml
rlorenzo: So we just need build and not image since the Dockerfile in build is already there?

rlorenzo:
@danpoltawski Okay, we didn't know that there was a set password.

But running a Behat test can be much different than running a manual test and there are times in which you want to see the test run to understand why it is failing.

Also, we cannot run the "And I pause" step using Docker. Any ideas on how to enable it?

abias:
> Why not use your own browser for that? I don't feel like using selenium node is best solution for this.

Let me rephrase that in accordance with Rex' use case:
I would really be interested in this feature for seeing what the Behat tests are doing visually and not only reading in the terminal that tests have passed.

base.yml
danpoltawski: Yep, the image is not being used, you either use an image or build.

But for the reasons I mention below, I don’t think you need to build anyway.

danpoltawski:
I've merged a slightly changed version of this in #49 - thanks a lot for the patch Rex

https://github.com/moodlehq/moodle-docker#using-vnc-to-view-behat-tests

PR #51, opened 2017-09-15
Move php version handling to an environment variable

### LINKED ISSUE #15: Reconsider branching strategy for versions
I am not sure if using branches of this project is the best way to manage php versions, it might be better to control it with an environment variable a bit like i have done with the browser. Needs some consideration.

danpoltawski:
The branching way of doing things made the repository harder to use and
also a lot of work maintaining changes between branches.

Using a variable is more versatile.

Fixes #15


scara:
Hi Dan,
great move 👍 😉 !
Looks like `db.mssql.7.0.yml` is missing as well as the Win counterpart in `moodle-docker-compose.cmd`.

HTH,
Matteo

danpoltawski:
> is missing as well as the Win counterpart 

Yep, it was WIP - added now, but I haven't tested it yet.

PR #59, opened 2017-09-19
webserver: Let Moodle hostname be configurable.

### LINKED ISSUE #33: Moodle mapping to localhost not always fine

scara:
Useful when the container host engine runs on a machine different
from your local dev environment. Defaults to 'localhost'.

Resolves #33.


danpoltawski:
Thanks Matteo, looks perfect!

PR #60, opened 2017-09-19
Support version notation for selenium browsers

danpoltawski:
With this patch, you can set MOODLE_DOCKER_BROWSER to options like the
existing options:
`firefox`
`chrome`
But also you can specify a tag/version for each, such as:
`firefox:2.53.1`
`chrome:3.5.2`

Fixes #20

Issue #20, opened 2017-07-12
Consider controlling selenium combinations per branch, or override...


stronk7:
Depending of the Moodle branches and browsers, there can be dependencies about which selenium/driver/browser [combo to use](https://docs.moodle.org/dev/Acceptance_testing/Browsers/Working_combinations_of_OS%2BBrowser%2Bselenium). For example 27_STABLE may require selenium 2.47.1 and current master 2.53.1.

While having a default (surely for master normally), is nice... being able to override it can be interesting too, specially when there is some important change. Imagine tomorrow we start using, for Firefox the new Marionette driver for 3.4 and up, that requires selenium 3. Previous versions should continue using selenium...

Not sure if that's simple env variable to be used (MOODLE_DOCKER_SELENIUM_VERSION) or something to be programmatically defined (matching the Docs) or what. But right now it's not possible to play with that variable.

Ciao :-)

danpoltawski:
Vaguely related to #15 too.

The thing is trying to control the complexity and keep it simple as possible, otherwise we may as well get people to edit their own docker compose configs.

danpoltawski:
My thought to achieve this is to extend the MOODLE_DOCKER_BROWSER variable to allow a optional`browser:versiontag` syntax.

So you could any of `firefox`, `chrome`, `chrome:3.5.2`, `firefox:2.53.1`

The ones without versions would default to our default version.

Any thoughts?

danpoltawski:
I have implemented this idea in https://github.com/moodlehq/moodle-docker/pull/60 - refinments welcome

stronk7:
Closed by #195, just merged.

PR #63, opened 2017-10-01
docker-compose Fixed bad quoting in variables under Windows

scara:
Without this fix the compose stops with an error:
>set MOODLE_DOCKER_WWWROOT=C:\Projects\Moodle\moodle-master
>set MOODLE_DOCKER_DB=mysql
>cd C:\Projects\Moodle\moodle-docker
>copy config.docker-template.php %MOODLE_DOCKER_WWWROOT%\config.php
>bin\moodle-docker-compose.cmd up -d
ERROR: no such image: selenium/standalone-firefox"":2.53.1: invalid reference format

Already found while working at #20 in refining #60.


danpoltawski:
Thanks Matteo

PR #65, opened 2017-10-09
Add instructions for phpunit to README.md + Align version numbers for behat instructions

abias:
Hi,

I noticed that there are no instructions for running phpunit tests in the README.md yet, so I just added them. Within this process, I aligned the version numbers for the behat instructions so that both example outputs in the instructions match.

Thanks,
Alex


danpoltawski:
Merged, thanks Alex

PR #71, opened 2017-12-04
Add support for locally hosted exttests

### LINKED ISSUE #70: Add support for external test server
Our tests are run against files located on http://download.moodle.org, however it would be advantageous to allow use fo the external tests hosted locally in some fashion.

This is already possible in the codebase, we just need to make use of it.

andrewnicols:
To avoid unnecessary dependency upon download.moodle.org, this makes use
of locally hosted external tests using an additional docker image.

This only adds HTTP external tests.
HTTPS external tests will require more in-depth changes due to use of
SSL certificates.

Fixes #70.

### DIFF
--- a/base.yml
+++ b/base.yml
@@ -20,6 +20,8 @@ services:
       POSTGRES_USER: moodle
       POSTGRES_PASSWORD: "m@0dl3ing"
       POSTGRES_DB: moodle
+  exttests:
+    image: moodlehq/moodle-exttests
   selenium:
     image: "selenium/standalone-firefox${MOODLE_DOCKER_SELENIUM_SUFFIX}:2.53.1"
     volumes:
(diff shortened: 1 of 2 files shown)


andrewnicols:
This change is incredibly simple, but is blocked by MDL-60948.

danpoltawski:
If previously blocked by upstream, why did the tests pass? Did we not have any external tests?

(In any case, seems like this could be merged)

andrewnicols:
It _was_ blocke dby upstream but it is no more.
The issue was that there was a test using getExternalTestUrl() twice - once with the HTTP flag, and once with HTTPS, and expecting the HTTP result to be converted into the HTTPS one by the httpsreplace tool. However the two values are independent and cannot be relied upon.

It should be safe to merge this now.

andrewnicols:
I've just confirmed that this now works as expected. Merged.

PR #74, opened 2018-01-08
Add possibility to use an alternative config file.

kabalin:
As a developer, I want to have an option to specify alternative config.php file, so that I can run tests on a code I am working currently without the need to modify existing config.php.

In reality, default template will work for user without modification, so in order to run unit test all developer needs to do is:
```
clone https://github.com/moodlehq/moodle-docker.git
cd moodle-docker
export MOODLE_DOCKER_WWWROOT=/path/to/moodle/code
export MOODLE_DOCKER_DB=pgsql
export MOODLE_DOCKER_CONFIGFILE=$(pwd)/config.docker-template.php
bin/moodle-docker-compose up -d
bin/moodle-docker-compose exec webserver php admin/tool/phpunit/cli/init.php
bin/moodle-docker-compose exec webserver vendor/bin/phpunit auth_manual_testcase auth/manual/tests/manual_test.php
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,11 +21,16 @@ This repository contains Docker configuration aimed at Moodle developers and tes
 ```bash
 # Set up path to Moodle code
 export MOODLE_DOCKER_WWWROOT=/path/to/moodle/code
+
 # Choose a db server (Currently supported: pgsql, mariadb, mysql, mssql, oracle)
 export MOODLE_DOCKER_DB=pgsql
 
-# Ensure customized config.php for the Docker containers is in place
-cp config.docker-template.php $MOODLE_DOCKER_WWWROOT/config.php
+# Ensure customized config.php for the Docker containers is in place. For
+# simple test run it should be OK using config.docker-template.php without
+# modifications by providing its location in MOODLE_DOCKER_CONFIGFILE env
+# variable. If MOODLE_DOCKER_CONFIGFILE is not defined,
+# $MOODLE_DOCKER_WWWROOT/config.php is used instead.
+export MOODLE_DOCKER_CONFIGFILE=$(pwd)/config.docker-template.php
 
 # Start up containers
 bin/moodle-docker-compose up -d
@@ -125,7 +130,8 @@ You can change the configuration of the docker images by setting various environ
 |-------------------------------------------|-----------|---------------------------------------|---------------|------------------------------------------------------------------------------|
 | `MOODLE_DOCKER_DB`                        | yes       | pgsql, mariadb, mysql, mssql, oracle  | none          | The database server to run against                                           |
 | `MOODLE_DOCKER_WWWROOT`                   | yes       | path on your file system              | none          | The path to the Moodle codebase you intend to test                           |
-| `MOODLE_DOCKER_PHP_VERSION`               | no        | 7.3, 7.2, 7.1, 7.0, 5.6                         | 7.1           | The php version to use                                                       |
+| `MOODLE_DOCKER_CONFIGFILE`                | no        | path on your file system              | config.php    | Alternative location for config.php to use                                   |
+| `MOODLE_DOCKER_PHP_VERSION`               | no        | 7.3, 7.2, 7.1, 7.0, 5.6               | 7.1           | The php version to use                                                       |
 | `MOODLE_DOCKER_BROWSER`                   | no        | firefox, chrome                       | firefox       | The browser to run Behat against                                             |
 | `MOODLE_DOCKER_PHPUNIT_EXTERNAL_SERVICES` | no        | any value                             | not set       | If set, dependencies for memcached, redis, solr, and openldap are added      |
 | `MOODLE_DOCKER_WEB_HOST`                  | no        | any valid hostname                    | localhost     | The hostname for web                                |
--- a/base.yml
+++ b/base.yml
@@ -6,6 +6,7 @@ services:
       - db
     volumes:
       - "${MOODLE_DOCKER_WWWROOT}:/var/www/html"
+      - "${MOODLE_DOCKER_CONFIGFILE}:/var/www/html/config.php"
       - "${ASSETDIR}/web/apache2_faildumps.conf:/etc/apache2/conf-enabled/apache2_faildumps.conf"
     environment:
       MOODLE_DOCKER_DBTYPE: pgsql
--- a/bin/moodle-docker-compose
+++ b/bin/moodle-docker-compose
@@ -74,6 +74,8 @@ then
     dockercompose="${dockercompose} -f ${basedir}/webserver.port.yml"
 fi
 
+# Alternative config.php
+export MOODLE_DOCKER_CONFIGFILE=${MOODLE_DOCKER_CONFIGFILE:-${MOODLE_DOCKER_WWWROOT}/config.php}
 
 # Mac OS Compatbility
 if [[ "$(uname)" == "Darwin" ]]; then


kabalin:
^ that check failure has nothing to do with this patch :)

danpoltawski:
This looks good to me 👍 

It might be worth talking about the alternative approach of adding a config.php hook where we load another file before including setup.php. That way, you could do your customisation in the other file, and not have to reimplement any new features that get added to the config.docker-template.php template. The catch of course is the constants.

Also, I suppose I could argue that at the moment, this project isn't doing anything to manage config.php at all, so could argue this option isn't necessary.

andrewnicols:
I would argue that we should not make this change.

I have also been thinking about the same change, but I discovered that docker is not well suited to mounting individual files rather than folders. In fact, I was watching/ reading about best practices and saw advice explicitly suggesting not to do this.

When I tried to do his myself some months ago I found other issues with it. Once the file is mounted it no longer accepts file changes made on the host OS for example; you cannot do things like replace the file with another.

Therefore I’d suggest that we do not follow through with this pull request in its current form.

As an alternative (a Nasty one), you could wrap both configs in an id test for one of the docker variables for example.

scara:
Hi @andrewnicols,
while I'm still thinking at the opportunity of this change, I'm posting about file mounting.

Even if you're not running on Windows, maybe https://github.com/docker/for-win/issues/328 and then https://github.com/moby/moby/issues/15793#issuecomment-135411504 could be of any help.

HTH,
Matteo

kabalin:
Thanks for response guys. I just came across explanation of the issue that @andrewnicols experienced before spotting @scara posted link to it, this explains why individual files updates are not propogated. But is it critical? Is there a chance you need to modify config.php while keeping containers running?

Well there is obviously an alternative way to tweak local config.php in a way that it would load certain file depending on environment variable, and in fact I used this method already by having a snippet below just before including setup.php:
```
if (getenv('MOODLE_DOCKER_DBTYPE')) {
    require_once($CFG->dirroot . '/config-test.php');
}
```
By suggesting this patch I was thinking more from the perspective to make this docker-based testing suit useful for quick test run out of box so to say.

andrewnicols:
I’ve tried to do this in the past and was going to push a pull request myself months ago.

The reason I haven’t is that it really messed with my workflow and didn’t work well at all. If you want to change anything in the config, it will not propagate well and it will be very confusing.
I’ve seen the docker community advise against it so it would get a -1 from me. What do others think?

scara:
Hi @kabalin,
just for the record, since you already said that you don't think anyone will be in need to make live changes on the config file:
> I used this method already by having a snippet below just before including setup.php:
>
> ```
> if (getenv('MOODLE_DOCKER_DBTYPE')) {
>     require_once($CFG->dirroot . '/config-test.php');
> }
> ```
this method won't work if we want to change something _live_, when running, since env vars are set **only** when starting up the process (ehm, the container :wink:).

HTH,
Matteo

kabalin:
@scara yes, env vars are set when container has been created and remain available in the container during its lifecycle. getenv('MOODLE_DOCKER_DBTYPE') is a bit misleading in example above, I am using it just as a flag that file is called inside the testing container and config-test.php is effectively a copy of config.docker-template.php from this repo.

kabalin:
Hello! Can this be re-considered for integration? I know at least 3 developers already who is using lucisgit/moodle-docker fork as it makes testing on dev box much simpler. Do I need to change something in code perhaps, so you can give it a green light for accepting? :)

scara:
Hi @kabalin,
I'm not part of Moodle HQ so mine is just a recommendation to let the PR be pretty ready: you should change `moodle-docker-compose.cmd` too.

HTH,
Matteo

kabalin:
Thanks @scara, I added a change to `moodle-docker-compose.cmd` as well.

kabalin:
Just to document here a discussion we had at some point with @andrewnicols .

> Ruslan Kabalin, [20.11.18 10:13]
> Hey Andrew, do I need to do something special to get https://github.com/moodlehq/moodle-docker/pull/74 merged?  ;)
> 
> Andrew Nicols, [20.11.18 13:36]
> As I said before I have concerns about it because it is recommended against by docker :-/
> 
> Ruslan Kabalin, [20.11.18 14:06]
> It is a hypothetical case that you need to modify config while container is running?
> 
> Ruslan Kabalin, [20.11.18 14:07]
> Also I am not convinced about "recommendation against by docker", as single file mounts is a common thing.
> 
> Ruslan Kabalin, [20.11.18 14:25]
> And in fact, there is no editing limitation of single file mounts which you describe, it seems like vim limitation in fact according to the issue Matteo refers to. I just tested modifying content using stdout redirect to file, and it propagates to inside running container without any problem 😁
> 
> Ruslan Kabalin, [20.11.18 16:05]
> Just thinking, I could probably re-write it using same approach Dan used for cached volume, i.e. if there is no custom config variable, it will not try to mount config.php separately. Will that resolve your main concern? ;)
> 
> Andrew Nicols, [20.11.18 23:42]
> That would be one solution yes. Another solution could be to bind mount to move the config to a new directory, bind-mount the entire directory, and symlink to config.php. You'd have to change the setup line to ../lib/setup.php
> 
> Andrew Nicols, [20.11.18 23:43]
> And people can have their local config in the same directory structure locally. The directory woudl be mounted over the top

andrewnicols:
Just wanted to note:
1. This is not a hypothetical. I, and many others, make config.php changes constantly and that includes after container start
1. Bind mounting of individual files is not recommended by docker
1. The suggested approach of creating a new config.php which uses an env variable to load an alternate file is probably the most pain free for all involved. i.e.
```
<?php
// Moodle configuration file.
unset($CFG);
global $CFG;

$CFG = new stdClass();

// Allow placement of an alternate config file in the .config directory.
// The filename relative to the .config can be specified.
$configfile = getenv('MOODLE_DOCKER_ALTERNATE_CONFIG');

if (empty($configfile)) {
    $configfile = "config.php";
}

$fqconfigfile = __DIR__ . "/.config/{$configfile}";

if (!is_readable($fqconfigfile)) {
    error_log("Unable to read {$fqconfigfile}. Aborting.");
    exit(1);
}

require_once($fqconfigfile);
require_once(__DIR__ . '/lib/setup.php');
```
You can remove this from the git index easily:
```
echo "/.config" >> .git/info/exclude
```

stronk7:
Hi,

I must confess that my very first thought is "why". Anybody wanting to use an alternate config.php file can copy it and done. Not sure which is the advantage of creating a new DOCKER variable to point to it. 

In my case, for example, I've a monster config.php file, backed by a configlib.php, able to switch between all databases and branches, able to run under both docker and local and changes to it are perfectly propagated to the container... and I don't need neither a file-mount neither a hook.

But, apart from that, surely I'd go for a hook (adding) better that a replacement + mount. What you want to add or how you want to organize it... is up to every developer. Can be a monolithic case (like mine), or one different config per database, or debugging/no debugging... or whatever.

So, from the 2 alternatives I've seen around, I think my fav is to just add a hook to ````config.docker-template.php```` and, those wanting to use it, have an env variable to say what wants to be included (being 100% clear that it's not a replacement, but an extension/addenda).

That way ppl and systems using moodle-docker won't need any modification (everything will continue working pretty much the same), and those interested into customizing the default config will have an extension (hook) mechanism to do so (mounted as part of the moodleroot directory, fixed, like Andrew comments above (.config/ dir), or completely free. I really don't mind much about that detail.

Thoughts?

kabalin:
Still not good for newcomers to run tests without `config.php` modifications. I was thinking more from perspective that one has a local system they use for development already, so config file already has settings people used to have for own deployment and configuration taste, and prefer to use `moodle-docker` for testing only rather than switching their whole workflow to it. Eloy and Andrew are probably right, at the end one may conditonally include  `config.docker-template.php` or just rename config file to `config.php_` when they run tests (and replace original file with copy of `config.docker-template.php`), and then back to original name when they finished with test run. Anyway closing this PR, but still using this patch locally 😃

PR #76, opened 2018-02-07
Fix env name for PHPUnit external services

mrmark:
I could be 100% wrong, but think there's a typo in the config file template.  This is my first time trying out the project, so I could easily be wrong.  Cheers!

### DIFF
--- a/config.docker-template.php
+++ b/config.docker-template.php
@@ -53,7 +53,7 @@ $CFG->behat_faildump_path = '/var/www/behatfaildumps';
 
 define('PHPUNIT_LONGTEST', true);
 
-if (getenv('MOODLE_DOCKER_PHPUNIT_EXTRAS')) {
+if (getenv('MOODLE_DOCKER_PHPUNIT_EXTERNAL_SERVICES')) {
     define('TEST_SEARCH_SOLR_HOSTNAME', 'solr');
     define('TEST_SEARCH_SOLR_INDEXNAME', 'test');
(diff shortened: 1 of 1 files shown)


scara:
Hi @mrmark,
that's wrong since `MOODLE_DOCKER_PHPUNIT_EXTRAS` is the env var set by the composer in [`phpunit-external-services.yml`](https://github.com/moodlehq/moodle-docker/blob/c7bd54cb37f57da7e96e9c8043a1256a5eec4553/phpunit-external-services.yml#L5).

That file is added in the _composition_ via the [`-f`](https://docs.docker.com/compose/reference/overview/#use--f-to-specify-name-and-path-of-one-or-more-compose-files) option by the env variable `MOODLE_DOCKER_PHPUNIT_EXTERNAL_SERVICES` in [`bin/moodle-docker-compose`](https://github.com/moodlehq/moodle-docker/blob/6d8bca22d2eb6e763d5142498aa4b2f410df66ed/bin/moodle-docker-compose#L62).

Shortly, there are two environment contexts:
1. the first and unique, created by you when you start the composition with `bin/moodle-docker-compose`
2. the second, potentially for each service, created by the composition and exposed one by one to the composed services

HTH,
Matteo

mrmark:
Ah, that's right.  That got me twice already, keep thinking those ENVs in the config file are from my host and not in the docker container.  Thanks for the quick reply.

PR #79, opened 2018-03-14
Allow multiple instances of Moodle simultaneously

jpahullo:
As seen on the issue #55 I share my simple proposal to face this problem.

I used the option from `docker-compose`:

```
-p, --project-name NAME     Specify an alternate project name (default: directory name)
```

In a very simple manner, we can build the whole stack as many times as we need. We only need to set up a different web port every time.

PR #82, opened 2018-03-15
Fixed CI build failure.

scara:
Fixed a failure in a GeoIP unit test, found while reviewing #79

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
     - "PHP=7.0 DB=mysql SUITE=behat BROWSER=firefox"
     - "PHP=7.0 DB=mariadb SUITE=behat BROWSER=firefox"
 install:
-    - git clone --branch v3.3.0 --depth 1 git://github.com/moodle/moodle $HOME/moodle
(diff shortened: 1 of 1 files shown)


scara:
Pinning the _pgsql_ version looks like harder than moving to the latest Moodle version.

Refs:
- https://github.com/travis-ci/travis-ci/issues/9346
- [MDL-61626](https://tracker.moodle.org/browse/MDL-61626)

scara:
Will update the PR by moving to `3.3.5` just released which runs smoothly with any known version of pgsql (MDL-61626).

stronk7:
+1 to move to just released. In fact surely could move to 3.4.2. I'm happy either way, TIA!

scara:
PR rebased.
@stronk7: we should stick w/ 3.3.x since here we're still testing against PHP 5.6 too.

HTH,
Matteo

stronk7:
> we should stick w/ 3.3.x since here we're still testing against PHP 5.6 too.

Good point, you're right, thanks!

PR #83, opened 2018-03-21
Unpin postgres:9.6.7

### LINKED ISSUE #78: Unpin postgres:9.6.7
Because of a problem with PG 9.6.8 (all builds failing consistently), we have had to pin PG to 9.6.7 everywhere. That has been done @ #77 .

So this is about to understand why it's failing (it maybe a real problem in Moodle, not in PG!).

And once the problem is fixed.... proceed to the previous situations (ping just to major 9).

stronk7:
This resolves #78, now that the driver has been fixed.

### DIFF
--- a/base.yml
+++ b/base.yml
@@ -15,7 +15,7 @@ services:
       MOODLE_DOCKER_BROWSER: firefox
       MOODLE_DOCKER_WEB_HOST: "${MOODLE_DOCKER_WEB_HOST}"
   db:
-    image: postgres:9.6.7
+    image: postgres:11
     environment:
       POSTGRES_USER: moodle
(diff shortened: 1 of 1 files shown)


stronk7:
BTW, warn! This will make 31 and 32 PG tests to be back to fail as far as [MDL-61626](https://tracker.moodle.org/browse/MDL-61626) was only for 32 and up.

So maybe we need to make PG sticky there, @ testing infra. For those 2 braches.

stronk7:
With Moodle <= 34_STABLE 100% out of support, now we can proceed with this. Also raising pg version to 11 (from 9). To run on the edge!

PR #89, opened 2018-11-26
Add node container support

mackensen:
This adds a node container, locked at the necessary core-supported version, for running grunt tasks. The script is based on a Stackoverflow post: https://stackoverflow.com/a/37102013. It's a work-around to ensure that the Moodle web root is loaded before the container runs the command. The expected result (verified locally) is that grunt watch is running and compiling changes. There's probably a more elegant solution to this but I wanted to share what I'd gotten working.

### DIFF
--- a/base.yml
+++ b/base.yml
@@ -26,3 +26,9 @@ services:
     image: "selenium/standalone-firefox${MOODLE_DOCKER_SELENIUM_SUFFIX}:2.53.1"
     volumes:
       - "${MOODLE_DOCKER_WWWROOT}:/var/www/html:ro"
+  node:
+    build:
+      context: ./
+      dockerfile: node.dockerfile
+    volumes:
+      - "${MOODLE_DOCKER_WWWROOT}:/var/www/html
--- a/node.dockerfile
+++ b/node.dockerfile
@@ -1,0 +1,13 @@
+FROM debian:stretch-slim
+
+RUN apt-get update && apt-get install -my gnupg curl
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
+
+RUN apt-get clean
+RUN apt-get update
+RUN apt-get install -y \
+        nodejs
+
+RUN /usr/bin/npm install -g grunt-cli
+ADD ./bin/start-grunt.sh .
+CMD ./start-grunt.sh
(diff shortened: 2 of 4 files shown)


node.dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM debian:stretch-slim
```
septatrix, line 1: Why not use `FROM node`? Also there is already a discussion about including node in the moodle-apache-php image

septatrix (changes requested):

base.yml
```diff
@@ -26,3 +26,9 @@ services:
     image: "selenium/standalone-firefox${MOODLE_DOCKER_SELENIUM_SUFFIX}:2.53.1"
     volumes:
       - "${MOODLE_DOCKER_WWWROOT}:/var/www/html:ro"
+  node:
+    build:
+      context: ./
+      dockerfile: node.dockerfile
+    volumes:
+      - "${MOODLE_DOCKER_WWWROOT}:/var/www/html
```
odeialba, line 34:
```suggestion
      - "${MOODLE_DOCKER_WWWROOT}:/var/www/html"
```

node.dockerfile
odeialba:
```suggestion
FROM node
```

node.dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+RUN apt-get update && apt-get install -my gnupg curl
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
```
odeialba, line 4:
```suggestion
RUN curl -sL https://deb.nodesource.com/setup_14.x | bash -
```

node.dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+RUN apt-get clean
+RUN apt-get update
+RUN apt-get install -y \
+        nodejs
+
+RUN /usr/bin/npm install -g grunt-cli
+ADD ./bin/start-grunt.sh .
```
odeialba, line 12:
```suggestion
ADD ./bin/start-grunt.sh .

USER root 
RUN chmod 755 ./start-grunt.sh

```

stronk7:
Hi,

as far as we include `.nvmrc` for all currently supported versions of Moodle (and older), wouldn't be better to include in the webserver image also nvm, so a simple nvm install & nvm use does the trick within the webserver container?

The problem I can see with this PR is that it requires us to be sticky to a given node version and that can be tricky.

Note I've not tried the idea above, maybe it has other problems, just I'm not sure if adding any extra node container really helps, hence, thinking about potential alternatives.

Also, side note: since #242 is possible to have a new `local.yml` file providing as many customisations as wanted. This may fall under those customisations easily, I imagine.

Thoughts about the idea (or any other idea related to ease of node/npm use) ?

Ciao :-)

PR #90, opened 2018-12-10
Set the path to python

dmonllao:


config.docker-template.php
```diff
@@ -82,4 +82,6 @@
     define('TEST_ENROL_LDAP_DOMAIN', 'ou=Users,dc=openstack,dc=org');
 }
 
+$CFG->pathtopython = 'python';
```
scara: Hi @dmonllao,
should be `/usr/bin/python` be more appropriate?

HTH,
Matteo

danpoltawski:
:+1: to full path

dmonllao:
:+1: thanks for the explanation Matteo. The patch was already updated using the full path so no further changes are required.

dmonllao:
Any chance that we could merge this? The patch does not seems problematic.

dmonllao:
The failures reported by travis are unrelated to this patch.

PR #92, opened 2018-12-26
Updated documentation for admin logging

adrienthiery:
Took me a bit to figure it out...


stronk7:
Thanks @adrienthiery ,

really the username defaulting to "admin" is a matter of ```admin/cli/install_database.php``` doing that, but it's not crazy to have it written down.

PR #98, opened 2019-03-22
Additional values for MOODLE_DOCKER_PHP_VERSION

rlorenzo:
I tried setting MOODLE_DOCKER_PHP_VERSION to 7.2 and it did not change the PHP version. Turns out I needed to set it to 7.2-stretch. So I added in additional values that MOODLE_DOCKER_PHP_VERSION can be used to include 7.2-stretch and 7.3-stretch. 

It is not obvious that this is the proper values to change PHP to a version higher than 7.1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ You can change the configuration of the docker images by setting various environ
 |-------------------------------------------|-----------|---------------------------------------|---------------|------------------------------------------------------------------------------|
 | `MOODLE_DOCKER_DB`                        | yes       | pgsql, mariadb, mysql, mssql, oracle  | none          | The database server to run against                                           |
 | `MOODLE_DOCKER_WWWROOT`                   | yes       | path on your file system              | none          | The path to the Moodle codebase you intend to test                           |
-| `MOODLE_DOCKER_PHP_VERSION`               | no        | 7.1, 7.0, 5.6                         | 7.1           | The php version to use                                                       |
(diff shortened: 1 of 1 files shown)


rlorenzo:
This is strange, now setting 7.2 via:

export MOODLE_DOCKER_PHP_VERSION=7.2-stretch
export MOODLE_DOCKER_PHP_VERSION=7.2

Do not work. I get PHP 7.0.32.

But the following work (in that I get the right PHP version):

export MOODLE_DOCKER_PHP_VERSION=7.2
export MOODLE_DOCKER_PHP_VERSION=7.1

What can be causing my setup to not load up PHP 7.2 images correctly?

andrewnicols:
Hi Rex,

It sounds like it's a local issue. The tags all exist for 7.1, 7.2, 7.3, 7.0, 5.6  (https://hub.docker.com/r/moodlehq/moodle-php-apache/tags). You should use those tags without the `-stretch` and `-jessie` suffixes. You _can_use the suffixes, but I would not recommend it and I do not think we should include that in the documentation either.

Maybe try `docker pull moodlehq/moodle-php-apache:7.2` and see if you can actually pull the image from the hub.

rlorenzo:
The docker pull command worked for me. I can now run PHP 7.2.

I updated my branch to add 7.2 and 7.3: https://github.com/rlorenzo/moodle-docker/commit/cd5185b2f75199a5fdbf20685dd0f3d2b9532d70

PR #103, opened 2019-06-20
Enable Redis session tests back for PHP 7.2 and up

stronk7:
(this is part of https://tracker.moodle.org/browse/MDLSITE-5876)

This reverts commit 2c7756fc9a51fade794dae67fe1bd780a4858c01.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,10 @@ services: docker
 fast_finish: true
 env:
     # Oracle jobs have been removed because there aren't public images anymore.
-    - "PHP=7.3 DB=pgsql   GIT=v3.7.0 SUITE=phpunit-full"
+    # We have set the "phpunit-full" to run against master because it's the
+    # weekly (3.8dev) where redis was fixed (MDL-60978). Once we have any working release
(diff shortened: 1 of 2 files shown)


stronk7:
Note this is failing the full run because we are running it against v3.7.0 that, obviously, doesn't have the fix applied. Once https://tracker.moodle.org/browse/MDLSITE-5876 is available @ moodle.git... we'll be able to change moodle-docker own travis tests to point to working moodle.git commit/version.

stronk7:
I've added a tiny change to travis configuration in order to get the "phpunit-full" job (the one running all the external tests, including redis) using current master, that includes the redis sessions fix. Once we have any release (v3.7.1 for example) with the fix applied... we'll be able to move back to tagged versions (or no, it's not strictly required to do so).

stronk7:
Yuppi, passing. So, ready to be merged and allow ppl using moodle-docker to run redis session tests again :-)

andrewnicols:
Merging :)

andrewnicols:
Bah - hit the wrong button and now i Have to wait for tests to pass.

PR #107, opened 2019-09-11
Unpin 3.5.2 and allow any new 3.x release to be used

stronk7:
With [MDL-66378](https://tracker.moodle.org/browse/MDL-66378) and, specially,
https://github.com/moodlehq/php-webdriver/pull/1
we can continue using Chrome >= 76 because we are enforcing the
use of the legacy JsonWire protocol (instead of the default W3C one).

So it has been agreed that we'll allow any 3.x new release to be used in
order to detect problems earlier (we always can pin again if needed).

### DIFF
--- a/selenium.chrome.yml
+++ b/selenium.chrome.yml
@@ -4,6 +4,6 @@ services:
     environment:
       MOODLE_DOCKER_BROWSER: chrome
   selenium:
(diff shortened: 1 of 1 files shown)


stronk7:
(a complete local run is being executed just to verify it all passes)

stronk7:
note that the tests above are failing because they use 3.7.0 release and it was not ready to use our forked instaclick/php-webdriver.

So I'm going to amend the tests to use a more recent version with our fork in use... coming soon...

PR #113, opened 2019-12-05
Add mlbackend_python tests

dmonllao:


base.yml
```diff
@@ -26,3 +26,7 @@ services:
     image: "selenium/standalone-firefox${MOODLE_DOCKER_SELENIUM_SUFFIX}:2.53.1"
     volumes:
       - "${MOODLE_DOCKER_WWWROOT}:/var/www/html:ro"
+  mlbackendpython:
+    image: "moodlehq/moodle-mlbackend-python:2.3.0-python3.7.5"
+    ports:
+      - "5000:5000"
```
scara: Hi @dmonllao,
1. you've missed to add a new dependency to the `webserver` service
1. you don't need to expose the TCP port `5000` unless for other purposes: when you use a compose, a dedicated network interconnects all the services within that compose i.e. `mlbackendpython:5000` should be already available without mapping the port on the docker host too

HTH,
Matteo

dmonllao:
Many thanks for the review Matteo. I've updated the patch, removing the port and moving the definition of the new image to phpunit-external-services.yml

stronk7:
I'm a little bit worried about how we are going to manage different versions here whenever they diverge between branches or when moodle-docker is used for versions < 3.8.0 ... uhm...

dmonllao:
any similar case in moodle-docker? (e.g. different pg/redis/ldap versions used for different moodle major branches). If MOODLE_DOCKER_WWWROOT must be set when the containers are initialised (https://github.com/moodlehq/moodle-docker/blob/master/bin/moodle-docker-compose#L4) we could modify bin/moodle-docker-compose to:
- Read the info in version.php
- Set up a hardcoded mapping array from Moodle major branches to latest tags in https://hub.docker.com/repository/docker/moodlehq/moodle-mlbackend-python
- Assign the mlbackend tag to an env variable MOODLE_DOCKER_MLBACKEND_PYTHON_TAG (I guess we could use something like https://docs.docker.com/compose/environment-variables/#the-env-file)
- Modify the external services yml file so the image attribute looks like this:
image: "moodlehq/moodle-mlbackend-python:${MOODLE_DOCKER_MLBACKEND_PYTHON_TAG}"

stronk7:
I'm not aware of any other image/external service needing such differentiation. If something... optionally, maybe selenium versions... but that's apart/exceptional.

Yes... we should be able to do that via version.php parsing (if the env is not manually set). From the up shell & cmd scripts. And keep a table of branches and versions maintained somehow.

Cannot imagine another way, right now... if anybody knows... please, share your wisdom here!

Ciao :-)

dmonllao:
I've coded the approach proposed above in a separate branch: https://github.com/dmonllao/moodle-docker/commit/e1afc7f30009053a26a86ade456971f3665a5e72. I've set it as a separate branch because:
- I have no idea how to code this in bin/moodle-docker-compose.cmd
- What happens with master needs to be discussed. If we set the MOODLE_DOCKER_MLBACKEND_PYTHON_TAG variable mlbackend_python tests will be skipped silently. If we want you to update moodle-docker every time there is a new major release I would suggest not setting the value for > 38 so that docker composer up errors and it can be updated.

stronk7:
OT, for the records (spaming all you, beloved moodlers):

Now the python mlbackend server images are being auto-built. More info @ https://github.com/moodlehq/moodle-docker-mlbackend-python/issues/1

Ciao :-)

scara:
Hi @stronk7,
> I have no idea how to code this in bin/moodle-docker-compose.cmd
is that under your radar?

I could try to do that but not in the very next days.

HTH,
Matteo

stronk7:
Yeah, I'll be looking later this week. Need to close it (integration-wise). I bet it's going to be tricky (but haven't had time to look to the login implemented in unix shell yet).

Will share here any progress (or lack of).

PR #66, opened 2017-10-10
Add env variables to control the ip address a port binds to

abias:
This PR solves issue #64.
With this patch, I decided to bind to localhost-only from now on. This is a change in the current default behavior, but it will improve the security of the docker containers.

PR #114, opened 2019-12-05
Don't expose web server externally

tmuras:
By default do not expose web server externally.


abias:
Hi @tmuras ,

I think your pull request is a duplicate of #66 which has been discussed in october with @stronk7 and which just would have to be finished by me or anybody else.

If you have the time to assist with #66, this would be highly appreciated and would also make this pull request here obsolete.

Cheers,
Alex

tmuras:
You are correct Alex, thanks.

PR #122, opened 2020-04-01
Mobile app support

NeillM:
Adds a MOODLE_MOBILE_VERSION environment parameter do define a
Moodle app docker image.

If the browser is chrome and it is defined the mobile app will be started and
Moodle configured so it is available for behat tests.

### DIFF
--- a/moodle-app.yml
+++ b/moodle-app.yml
@@ -4,4 +4,4 @@ services:
     environment:
       MOODLE_DOCKER_MOBILE: "true"
   mobileapp:
-    image: "moodlehq/moodlemobile2:${MOODLE_MOBILE_VERSION}"
+    image: "moodlehq/moodleapp:${MOODLE_MOBILE_VERSION}"
--- a/moodle-app.yml
+++ b/moodle-app.yml
@@ -1,0 +1,7 @@
+version: "2"
+services:
+  webserver:
+    environment:
+      MOODLE_DOCKER_MOBILE: "true"
+  mobileapp:
+    image: "moodlehq/moodlemobile2:${MOODLE_MOBILE_VERSION}"
--- a/moodle-app.yml
+++ b/moodle-app.yml
@@ -2,6 +2,6 @@ version: "2"
 services:
   webserver:
     environment:
-      MOODLE_DOCKER_MOBILE: "true"
-  mobileapp:
-    image: "moodlehq/moodleapp:${MOODLE_MOBILE_VERSION}"
+      MOODLE_DOCKER_APP: "true"
+  moodleapp:
+    image: "moodlehq/moodleapp:${MOODLE_APP_VERSION}"
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ install:
     - export MOODLE_DOCKER_BROWSER=$BROWSER
     - export MOODLE_DOCKER_WWWROOT="$HOME/moodle"
     - export MOODLE_DOCKER_PHP_VERSION=$PHP
-    - export MOODLE_MOBILE_VERSION=$MOBILE
+    - export MOODLE_APP_VERSION=$MOBILE
 before_script:
     - tests/setup.sh
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,9 @@ env:
     - "PHP=7.1 DB=mariadb GIT=v3.8.0 SUITE=behat BROWSER=firefox"
     - "PHP=7.1 DB=mariadb GIT=v3.7.2 SUITE=behat BROWSER=chrome"
     - "PHP=7.0 DB=mariadb GIT=v3.5.8 SUITE=behat BROWSER=firefox"
+    # Mobile app
+    - "PHP=7.3 DB=pgsql GIT=master SUITE=app BROWSER=chrome MOBILE=next"
+    - "PHP=7.3 DB=pgsql GIT=master SUITE=app BROWSER=chrome MOBILE=latest"
 install:
     - git clone --branch $GIT --depth 1 git://github.com/moodle/moodle $HOME/moodle
     - cp config.docker-template.php $HOME/moodle/config.php
@@ -62,6 +65,7 @@ install:
     - export MOODLE_DOCKER_BROWSER=$BROWSER
     - export MOODLE_DOCKER_WWWROOT="$HOME/moodle"
     - export MOODLE_DOCKER_PHP_VERSION=$PHP
+    - export MOODLE_MOBILE_VERSION=$MOBILE
 before_script:
     - tests/setup.sh
 script:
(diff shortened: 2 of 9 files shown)


scara:
Hi @NeillM,
TNX for your contribution.

Is there any reason for the small difference between the new ENV VARs check in Linux vs in Windows?

TIA,
Matteo

NeillM:
I was pretty sure they were the same (or at least have the same result) i.e. to only run the app if the variable is set and the browser is Chrome.

If it is a better way it is entirely due to my ignorance 😄

scara:
Hi @NeillM,
just my curiosity given that the result will be the same:
``` bash
if [[ ! -z "$MOODLE_DOCKER_BROWSER" ]] && [[ "$MOODLE_DOCKER_BROWSER" == "chrome" ]] && [[ ! -z "$MOODLE_MOBILE_VERSION" ]];
```
vs
``` bash
IF "%MOODLE_DOCKER_BROWSER%"=="chrome" (
    IF NOT "%MOODLE_MOBILE_VERSION%"=="" (
```
i.e. the unbalancing is in `[[ ! -z "$MOODLE_DOCKER_BROWSER" ]]`, I mean why checking for being not empty if we need an exact value: that would be my ignorance too 😉 !

HTH,
Matteo

NeillM:
I'm not certain how useful the two tests I added actually will be (since core does not have any Moodle app tests) but I think they will find out if there is a horrible error starting the app or not.

NeillM:
Looks like the next test failed (the mobile app url was inaccessible for some reason) trying a rerun of it.

I guess the app might need a longer wait to start than some of the other images.

NeillM:
Lets see if adding a wait for app to be up script gets everything passing. Making it executable on Linux from Windows was fun.

NeillM:
It looks like adding a script to wait for the app to finish loading is just what was needed.

moodle-app.yml
```diff
@@ -0,0 +1,7 @@
+version: "2"
+services:
+  webserver:
+    environment:
+      MOODLE_DOCKER_MOBILE: "true"
+  mobileapp:
+    image: "moodlehq/moodlemobile2:${MOODLE_MOBILE_VERSION}"
```
crazyserver, line 7: Change it to moodlehq/moodleapp since moodlemobile2 is deprecated.

crazyserver:
+1 for me on the integration of that issue. But first the image name should be changed.

NeillM:
I have updated the app image name.

I'm glad the docs for the app repository name got updated as well.

NeillM:
hmm, it looks like the Travis tests are failing with the error:

> [RuntimeException]
>  Failed to clone https://github.com/moodlehq/php-webdriver.git, git was not found, check that it is installed and in your PATH env.  
>
>  sh: 1: git: not found

When Moodle is trying to install the test environment, pretty sure that is nothing to do with this change.

stronk7:
> hmm, it looks like the Travis tests are failing with the error..

Yeah that happens here and there because 1) composer hits the github API limits (for vcs-based stuff), then tries to download via git, but the moodle php images are git-less. And we have vcs-based repo for that package because the upstream one is broken and we had to fork it and make it look like the original.

Locally there are a number of workarounds (install git in local images, install composer in host, provide a token...) but from .travis it's not so easy, AFAIK.

So I bet that will end adding some token (if we can keep it hidden) and/or adding git to the images so that alternative install method is available when Github's API limits happen.

stronk7:
Just FYI, we are rebuilding all the `moodle-php-apache` images to include git. That will help all those tests to pass and also will allow some other dev workflows. Will be re-running the tests while new images become available...

stronk7:
Ah, confirmed, now the tests, with git available in the images, are passing. Will look to this later today, thanks for your patience!

scara:
> Just FYI, we are rebuilding all the moodle-php-apache images to include git.

Refs: https://github.com/moodlehq/moodle-php-apache/pull/110, https://github.com/moodlehq/moodle-php-apache/pull/111, https://github.com/moodlehq/moodle-php-apache/pull/112, https://github.com/moodlehq/moodle-php-apache/pull/113, https://github.com/moodlehq/moodle-php-apache/pull/115.

HTH,
Matteo

stronk7:
Hi guys,

I've been using this to test [MDL-68789](https://tracker.moodle.org/browse/MDL-68789) and I'm having some problems:

1. Under windows, with local_moodlemobileapp installed, I run the {{--tags=@app}} and it's passing perfectly: 20 scenarios (20 passed)
2. But with the mac, also with local_moodlemobileapp installed, I run it and I'm getting all the time: No scenarios / No steps

I've started from scratch to guarantee that everything was in place:

```
MOODLE_DOCKER_WWWROOT=/Users/stronk7/git_moodle/ci-tests
MOODLE_DOCKER_DB=pgsql
MOODLE_DOCKER_BROWSER=chrome
MOODLE_MOBILE_VERSION=next
```

Same result. Can you imagine anything? Sure it's one of those idiot details, but right now I'm lost.

Ciao :-)

stronk7:
Oh, lalala. I forgot to copy the new `config.docker-template.php`, lalala! Thanks for the rubber-ducking!

stronk7:
Ok, passed here without problem, both from mac and windows hosts (plus travis covers linux).

Only little detail is if we should also change other occurrences of "mobile" to "app", like:

- MOODLE_MOBILE_VERSION => MOODLE_APP_VERSION
- MOODLE_DOCKER_MOBILE => MOODLE_DOCKER_APP
- MOBILE => APP
- mobileapp => moodleapp

Thoughts? I'm not super strong on it, just find it better if we always name it the very same (app instead of mobile)

Ciao :-)

crazyserver:
+1 to your suggestion on the APP naming.

NeillM:
Sounds like a reasonable change, I should be able to get around to it soon.

stronk7:
The changes look perfect, @NeillM .

I'm just trying to be able to perform a run of @app to verify that everything is ok and for some reason, right now, I'm unable to achieve that.

With moodle-docker it seems that the app container dies, with standalone docker/moodle-ci-runner I'm getting a lot of failures, see https://github.com/moodlehq/moodle-ci-runner/pull/26  ... will try later, right now I'm out of ideas!

stronk7:
This is completely weird, I just did, to cleanup everything:
```
docker system prune -a -f --volumes
```
also restarted the docker service... and now moodleapp is up, running and passing:
```
$ bin/moodle-docker-compose exec -u www-data webserver php admin/tool/behat/cli/run.php --tags=@app
Running single behat site:
Moodle 3.9dev+ (Build: 20200527), a4bd55725b8c64f11ee4aa7fab888c51c13e0bd4
Php: 7.2.31, pgsql: 11.8 (Debian 11.8-1.pgdg90+1), OS: Linux 4.19.76-linuxkit x86_64
Server OS "Linux", Browser: "chrome"
Browser specific fixes have been applied. See http://docs.moodle.org/dev/Acceptance_testing#Browser_specific_fixes
Started at 29-05-2020, 01:54
...................................................................... 70
...................................................................... 140
...................................................................... 210
...................................................................... 280
...................................................................... 350
...................................................................... 420
...................................................................... 490
......................................................

20 scenarios (20 passed)
544 steps (544 passed)
19m4.70s (73.73Mb)
```

So, before moodleapp changes again... about to merge this now!

stronk7:
Merged, thanks!

stronk7:
I really don't understand what's going with the ionic docker image, since the run above yesterday, I've been unable to complete any other run (@ https://github.com/moodlehq/moodle-ci-runner/pull/26).

- I've tried pruning everything and restarting docker.
- I've tried increasing the behat timeout to `$CFG->behat_increasetimeout = 2;` still getting lots of AJAX/JS timeouts (right now I'm trying with the timeout set to 5 (that means 50 seconds!).
- Sometimes the image does not start, exiting in the middle of the startup.

Note it's not related with this issue, just sharing in case you can imagine any reason. My laptop is basically idle and only running those attempts (and it's a brand-new rmbp, plenty of processors and ram).

And for some reason, those ionic images (have tried both "next" and "latest") are really slow here, grrr.

Ciao :-)

stronk7:
This is what I'm getting with the crazy 50 seconds timeout (no more AJAX/JS timeouts, but still a good number of failures): https://pastebin.com/NVvFH5WP

NeillM:
I'm afraid I haven't been doing anything with the Moodle app for a while

It turned out our automatic test server could not handle having the app image loaded... we started getting random fails in our tests when it was up, not just the app ones, so it hasn't been running them for a good while now.

stronk7:
Thanks @NeillM !

NeillM:
I have turned them on again and the app tests all passed on the latest app image, using a recent Moodle 3.8

PR #126, opened 2020-06-10
Moodle app development

NoelDeMartin:
The current configuration runs the mobile app from docker images, and so it doesn't support using this for development.

With the modifications made in this PR, it's possible to configure a path to a local clone using the env variable `MOODLE_DOCKER_APP_PATH` (in the same way that a local clone of moodle is configured using `MOODLE_DOCKER_WWWROOT`).

~When I looked into the configuration I noticed that the previous env variable was called `MOODLE_APP_VERSION`, so I renamed it to `MOODLE_DOCKER_APP_VERSION` to keep the `MOODLE_DOCKER_` prefix. This is a breaking change, but considering that the variable was added recently it shouldn't affect too many developers. If this change is deemed too risky, I can undo it.~ --> Moved to a dedicated issue: #127.

I also noticed that app tests in Travis were running against the `next` image of the mobile app. I removed it because that image is too unstable (it's the development version). It may cause this project's CI to break if the development version's CI breaks, which isn't uncommon in a development branch. Testing with stable branches should be enough, like it's done for LMS tests.

~Finally, I had to bump the version of composer config to 2.3 in order to declare the [healthcheck](https://docs.docker.com/compose/compose-file/compose-file-v2/#healthcheck) in the app development config.~ --> Will be open in a different PR.

### DIFF
--- a/moodle-app-dev.yml
+++ b/moodle-app-dev.yml
@@ -1,0 +1,19 @@
+version: "2"
+services:
+  webserver:
+    environment:
+      MOODLE_DOCKER_APP: "true"
+  moodleapp:
+    image: node:11
+    working_dir: /app
+    command: npm run ionic:serve
+    volumes:
+      - "${MOODLE_DOCKER_APP_PATH}:/app"
+    expose:
+      - 8100
+      - 35729
+      - 53703
+    ports:
+      - "8100:8100"
+      - "35729:35729"
+      - "53703:53703"
--- a/tests/setup.sh
+++ b/tests/setup.sh
@@ -12,6 +12,15 @@ elif [ "$SUITE" = "phpunit-full" ];
 then
     export MOODLE_DOCKER_PHPUNIT_EXTERNAL_SERVICES=true
     initcmd="bin/moodle-docker-compose exec -T webserver php admin/tool/phpunit/cli/init.php"
+elif [ "$SUITE" = "behat-app-development" ];
+then
+    git clone --branch "v$APP_VERSION" --depth 1 git://github.com/moodlehq/moodleapp $HOME/app
+    git clone --branch "v$APP_VERSION" --depth 1 git://github.com/moodlehq/moodle-local_moodlemobileapp $HOME/moodle/local/moodlemobileapp
+
+    docker run --volume $HOME/app:/app --workdir /app node:11 npm install
+    docker run --volume $HOME/app:/app --workdir /app node:11 npm run setup
+
+    initcmd="bin/moodle-docker-compose exec -T webserver php admin/tool/behat/cli/init.php"
 elif [ "$SUITE" = "behat-app" ];
 then
     git clone --branch "v$APP_VERSION" --depth 1 git://github.com/moodlehq/moodle-local_moodlemobileapp $HOME/moodle/local/moodlemobileapp
--- a/tests/setup.sh
+++ b/tests/setup.sh
@@ -12,8 +12,10 @@ elif [ "$SUITE" = "phpunit-full" ];
 then
     export MOODLE_DOCKER_PHPUNIT_EXTERNAL_SERVICES=true
     initcmd="bin/moodle-docker-compose exec -T webserver php admin/tool/phpunit/cli/init.php"
-elif [ "$SUITE" = "app" ];
+elif [ "$SUITE" = "behat-app" ];
 then
+    git clone --branch "v$APP_VERSION" --depth 1 git://github.com/moodlehq/moodle-local_moodlemobileapp $HOME/moodle/local/moodlemobileapp
+
     initcmd="bin/moodle-docker-compose exec -T webserver php admin/tool/behat/cli/init.php"
 else
     echo "Error, unknown suite '$SUITE'"
--- a/tests/test.sh
+++ b/tests/test.sh
@@ -12,9 +12,9 @@ then
 elif [ "$SUITE" = "phpunit-full" ];
 then
     testcmd="bin/moodle-docker-compose exec -T webserver vendor/bin/phpunit --verbose"
-elif [ "$SUITE" = "app" ];
+elif [ "$SUITE" = "behat-app" ];
 then
-    testcmd="bin/moodle-docker-compose exec -T webserver php admin/tool/behat/cli/run.php --tags=@app"
+    testcmd="bin/moodle-docker-compose exec -T webserver php admin/tool/behat/cli/run.php --tags='@app&&@mod_login'"
 else
     echo "Error, unknown suite '$SUITE'"
     exit 1
--- a/tests/test.sh
+++ b/tests/test.sh
@@ -12,9 +12,9 @@ then
 elif [ "$SUITE" = "phpunit-full" ];
 then
     testcmd="bin/moodle-docker-compose exec -T webserver vendor/bin/phpunit --verbose"
-elif [ "$SUITE" = "behat-app" ];
+elif [ "$SUITE" = "behat-app" ] || [ "$SUITE" = "behat-app-development" ];
 then
-    testcmd="bin/moodle-docker-compose exec -T webserver php admin/tool/behat/cli/run.php --tags='@app&&@mod_login'"
+    testcmd="bin/moodle-docker-compose exec -T webserver php admin/tool/behat/cli/run.php --tags=@app&&@mod_login"
 else
     echo "Error, unknown suite '$SUITE'"
     exit 1
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,8 @@ env:
     - "PHP=7.1 DB=mariadb GIT=v3.7.2 SUITE=behat BROWSER=chrome"
     - "PHP=7.0 DB=mariadb GIT=v3.5.8 SUITE=behat BROWSER=firefox"
     # Mobile app
-    - "PHP=7.3 DB=pgsql GIT=master SUITE=behat-app BROWSER=chrome APP=3.9.0"
+    - "PHP=7.3 DB=pgsql GIT=master SUITE=behat-app-development BROWSER=chrome APP_VERSION=3.9.0 APP_PATH=$HOME/app"
+    - "PHP=7.3 DB=pgsql GIT=master SUITE=behat-app BROWSER=chrome APP_VERSION=3.9.0"
 install:
     - git clone --branch $GIT --depth 1 git://github.com/moodle/moodle $HOME/moodle
     - cp config.docker-template.php $HOME/moodle/config.php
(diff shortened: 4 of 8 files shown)


NeillM:
I would prefer the environment variable don't get name changed (I would have to go and fix our automatic testing jobs if it happens)

moodle-app-dev.yml
```diff
@@ -0,0 +1,24 @@
+version: "2.3"
```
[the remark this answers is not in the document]
NoelDeMartin: Ok, I'll move to to a different PR.

NoelDeMartin:
@scara How do you see it after the modifications?

stronk7:
Why are we changing the tags from `@app` to `@behat-app` and `@behat-app-development` ? Has that been documented anywhere? People out there using behat testing in their plugins...

tests/setup.sh
```diff
@@ -12,8 +12,19 @@ elif [ "$SUITE" = "phpunit-full" ];
 then
     export MOODLE_DOCKER_PHPUNIT_EXTERNAL_SERVICES=true
     initcmd="bin/moodle-docker-compose exec -T webserver php admin/tool/phpunit/cli/init.php"
-elif [ "$SUITE" = "app" ];
+elif [ "$SUITE" = "behat-app" ];
 then
+    git clone --branch v3.8.2 --depth 1 git://github.com/moodlehq/moodle-local_moodlemobileapp $HOME/moodle/local/moodlemobileapp
```
stronk7, line 17: Shouldn't this be using $APP_VERSION instead of harcoded `v3.8.2`?

tests/setup.sh
```diff
@@ -12,8 +12,19 @@ elif [ "$SUITE" = "phpunit-full" ];
...
+
+    initcmd="bin/moodle-docker-compose exec -T webserver php admin/tool/behat/cli/init.php"
+elif [ "$SUITE" = "behat-app-development" ];
+then
+    git clone --branch v3.8.2 --depth 1 git://github.com/moodlehq/moodleapp $HOME/app
```
stronk7, line 22: And, for testing purposes, no matter it's not used normally... maybe we could also pass $APP_VERSION (just thinking that we may want to test multiple or, tomorrow... we can change the versions without touching the scripts.

tests/test.sh
```diff
@@ -12,9 +12,9 @@ then
 elif [ "$SUITE" = "phpunit-full" ];
 then
     testcmd="bin/moodle-docker-compose exec -T webserver vendor/bin/phpunit --verbose"
-elif [ "$SUITE" = "app" ];
+elif [ "$SUITE" = "behat-app" ] || [ "$SUITE" = "behat-app-development" ];
 then
-    testcmd="bin/moodle-docker-compose exec -T webserver php admin/tool/behat/cli/run.php --tags=@app"
+    testcmd="bin/moodle-docker-compose exec -T webserver php admin/tool/behat/cli/run.php --tags='@app&&@mod_forum&&@app_from3.7'"
```
stronk7, line 17: Why is this leading to an execution with ZERO scenarios in the [travis build](https://travis-ci.org/github/moodlehq/moodle-docker/builds/703572428)? See the last 2 jobs, that are the ones introduced by this PR.

tests/setup.sh
NoelDeMartin: Yes that's a good improvement, I've added it :).

tests/setup.sh
NoelDeMartin: I couldn't do this before because of variables precedence (`MOODLE_APP_VERSION` took precedence over `MOODLE_DOCKER_APP_PATH`). I've changed the precedence to support this in the tests. It shouldn't be a problem outside of travis because the two variables are not supposed to be used together.

tests/test.sh
NoelDeMartin: I'm not sure why, maybe I did something wrong the last time. In any case, I've updated the code to use 3.9.0 and some tests are being run now.

NoelDeMartin:
@stronk7 Thanks for taking a look!

I've applied your suggestions and updated some of the code with more recent versions (3.9.0 was released after I opened this PR).

> I only have an "existential" question, about how/why/when this is supposed to work when the local moodleapp is used. I'm really failing to find its utility within moodle-docker, and I'm curious about it.

I've added a new section in the README talking about this. In a nutshell, the idea is that this is useful to run Behat tests for the mobile app because they require a Moodle environment. For development, I think it's better to work with a local installation of npm. Although this is also nice for doing TDD using Behat.

tests/setup.sh
stronk7: Good!

tests/setup.sh
stronk7: Yeah, saw that if/elif precedence later. Good fix!

stronk7:
Thanks for all the changes, @NoelDeMartin . Here there are 3 more points, about the latest iteration (I won't create a new review because it's confusing for me, heh).

1) I'm not fully convinced about the "harcoded" `v` [here](https://github.com/moodlehq/moodle-docker/pull/126/files#diff-89d361c802359923d1628b500e1c3682R17) (and in a few other places). If the version is `v3.9.0`, then that's the version and the tag, and should be passed that way. It's the 1st time I see that, I think.

2) About the [added instructions](https://github.com/moodlehq/moodle-docker/pull/126/files#diff-04c6e90faac2675aa89e2176d2eec7d8R102)... I think we must be super clear about when you just want to run the tests of your plugin (using `MOODLE_APP_VERSION`) using the fully dockered moodleapp or when you want to use a local (`MOODLE_DOCKER_APP_PATH `) moodleapp (also just for running tests or maybe also to hack the app itself. When I've read the changes in the README, initially I thought it was the former ("Use containers for running behat tests") but after reading I realised that it was the later. For the newcomer it will be way more confusing, I bet.

3) Already detected by you... it seems that travis new jobs using the 2 app alternatives... still get zero scenarios run. Strange.

Ciao :-)

NeillM:
When I added the first app test core Moodle did not include any app tests (I'm assuming it still does not).

I did find it was still valuable to run as behat tests that the app is available when it starts up; so there will be an error in them if the app does not start properly.

stronk7:
> When I added the first app test core Moodle did not include any app tests (I'm assuming it still does not).
> 
> I did find it was still valuable to run as behat tests that the app is available when it starts up; so there will be an error in them if the app does not start properly.

Yeah, but with this PR, we are always installing https://github.com/moodlehq/moodle-local_moodlemobileapp that, for sure, contains `@app` tests. That's why the zero scenarios result is "strange" (unexpected).

NeillM:
To me it looks as though by default the plugin adds no feature files to Moodle, i.e. there are none in tests/behat

It kind of looks as though you would need to copy the contents of it's mod folder into Moodle's mod folder for that.

stronk7:
But the plugin ([local_moodlemobileapp](https://github.com/moodlehq/moodle-local_moodlemobileapp/)) that now is installed, adds lots of features, including that `@mod_login` one ([link](https://github.com/moodlehq/moodle-local_moodlemobileapp/blob/master/mod/login/tests/behat/app_basic_usage.feature#L1)] being set in the tests (where or not that "mod_login" tag is correct is another thing, but the feature and scenarios are there, I'd expect them to be run).

Sure I'm missing something, but cannot imagine what!

NeillM:
I didn't think behat would look inside /local/moodlemobileapp/mod/... for feature files

stronk7:
> I didn't think behat would look inside /local/moodlemobileapp/mod/... for feature files

Ah, I now see your point... but... I think it does... (though I'm not 100% sure). Behat's "finder" from memory is not filtering (like phpunit does). Trying here...

stronk7:
> > I didn't think behat would look inside /local/moodlemobileapp/mod/... for feature files
> 
> Ah, I now see your point... but... I think it does... (though I'm not 100% sure). Behat's "finder" from memory is not filtering (like phpunit does). Trying here...

It worked locally, and those @mod_login scenarios were found and run: https://pastebin.com/BvrwgrXb

NoelDeMartin:
I've finally found what's going on! As it often happens, it turned out to be something silly.

It seems like this line in the `test.sh` script:

```sh
testcmd="bin/moodle-docker-compose exec -T webserver php admin/tool/behat/cli/run.php --tags='@app&&@mod_login'"
```

should be this instead:

```sh
testcmd="bin/moodle-docker-compose exec -T webserver php admin/tool/behat/cli/run.php --tags=@app&&@mod_login"
```

I suppose there is something going on that eludes my bash scripting knowledge. I also tried using `--tags=\"@app&&@mod_login\"` to no avail. There must be something I don't know about how bash handles strings for executing commands, because if you run that in a terminal it will interpret the `&&` as an instruction to run two separate commands. That's why I was quoting the value to the `tags` argument. But it doesn't seem to be necessary.

@stronk7 responding to your comments:

1. I'm not convinced with this approach either, but the versions in docker hub are tagged like `3.9.0` whereas the versions in github are tagged like `v3.9.0`. The only way to avoid doing this would be using two different env variables for docker hub and github or removing the `v` from the env variable, which would probably be uglier than adding the `v`. I think the current approach is a good compromise.

2. Ok, I have updated the explanation mentioning both scenarios. Let me know if it's more understandable now.

3. As per my comments above, this should be fixed now.

stronk7:
Thanks @NoelDeMartin ,

Amazing the "quotes-gate" specially when the work ok locally (none, single and double). Eagle eye!

It's all ok for me, the only tiny detail that I can see is that `MOODLE_DOCKER_BROWSER` set to "chrome" is mentioned only when using the local moodleapp, when it's required also for the fully dockered alternative.

Other than that, I think this is pretty ready to go, yay!

stronk7:
(and OT and not needed for this PR... but I'd recommend to go normalizing the version names in github & docker, it's a little bit strange to have them different, plus current status doesn't allow to add (own) tests to moodle-docker by, for example, using "stable" or "next", we can use only numeric ones, that's the proof it's not a good situation).

NoelDeMartin:
Thanks for catching that, it's fixed in the README.

About the version naming, I agree it's a bit weird but it's a matter of conventions. In github, it's the convention to add the `v` and in docker hub it's the convention to avoid it. For example, we use the `v` in [moodle/moodle](https://github.com/moodle/moodle/releases) in github and we don't have it on any of [our images](https://hub.docker.com/u/moodlehq) in docker hub. I guess the correct solution would be to have two variables, because named versions can also be different (for example, there may be a "stable" tag in docker hub that is called "master" in github).

I think it's ok to leave it like this for now because it only affects the test scripts, users of `moodle-docker` will always use the version in docker hub and it works using named versions as well.

stronk7:
> I think it's ok to leave it like this for now because it only affects the test scripts, users of `moodle-docker` will always use the version in docker hub and it works using named versions as well.

Yeah, I agree, it's only for testing. Still, if tomorrow I'd so a moodlehq/moodle docker image... I think I'd use vX.Y.Z. Imagine the name are not digits, but we call it "Hot Sauce"... name (tag) should be the same, using different ones is "weird" (at very least to my "square-brain".

Anyway... for another day... doing the last review... and hopefully... landing this... great work!

stronk7:
Here we go, Happy Birthday, @NoelDeMartin (now I can pretend I delayed this on purpose for the celebrations, ROFL).

Thanks!

PR #81, opened 2018-03-15
add suport for .env files

jpahullo:
This is also really helpful since you do not need to `export` variables and have side effects when running several instances.

PR #130, opened 2020-06-23
Use .env file for directory/db

### LINKED ISSUE #80: Add suport for .env files
docker-compose suport `.env` files to load environment variables.

Since the `bin/moodle-docker-compose` script expects two environment variables already set up, it makes impossible to let docker-compose use the content of the `.env` for value replacing on the configuration loaded from YML files.

**Proposal:**

I think it is a very easy step to check if the file exists on the `/` directory, and load all configuration from the `.env` file before guard conditions start.

iwasherefirst2:
Using an environment file for directory/db
has the advantage that the user does not need
to specify the variables after each reboot.

Added `.env` to gitignore list, to avoid
that environment variables end up in repository.

Created `moodle-start` for convenience.

Updated readme.

Closes #80
Closes #81


README.md
```diff
@@ -18,27 +18,15 @@ This repository contains Docker configuration aimed at Moodle developers and tes
...
 ## Quick start
 
-```bash
-# Set up path to Moodle code
-export MOODLE_DOCKER_WWWROOT=/path/to/moodle/code
-# Choose a db server (Currently supported: pgsql, mariadb, mysql, mssql, oracle)
-export MOODLE_DOCKER_DB=pgsql
-
-# Ensure customized config.php for the Docker containers is in place
-cp config.docker-template.php $MOODLE_DOCKER_WWWROOT/config.php
```
danpoltawski, line 28: This still seems necessary for someone to quick start?

bin/moodle-start
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+bin/moodle-docker-compose up -d
+bin/moodle-docker-wait-for-db
```
danpoltawski, line 4: 1. I'm not very convinced of the value of this `moodle-start` script. My '_philosophy_' when originally designing this project was just to add the 'minimum helpers' to get by for convenience and scriptability, but encourage people to use the docker compose commands when necessary. This seems to abstracting what's going on under the hood more than I think this project should. For most people, doing `moodle-docker-compose up -d` will be enough for their workflow of development.

2. I don't think this script is well named.

danpoltawski:
I realise I just posted two negative comments - I think the rest of the `.env` support is a great idea - thank you for contributing it. :-) :+1:  - just wanted to make my thinking clear, @moodlehq now maintain this project.

README.md
iwasherefirst2: Whoups your right. I added a `create-moodle-config` script and updated the readme in the latest commit.

bin/moodle-start
iwasherefirst2: I understand your philosophy and removed the script.

iwasherefirst2:
> I realise I just posted two negative comments - I think the rest of the `.env` support is a great idea - thank you for contributing it. :-) +1 - just wanted to make my thinking clear, @moodlehq now maintain this project.

Thank you. I also think using an `.env` is very convenient. Writing the path to your Moodle in the terminal every time after I boot my computer was time-consuming (long path) and error prone (typo). Also when I read the readme for the first time,  I was very confused about the `Quick start` section. I was sure that this was a script that I have to adjust, so I went to all files in the folder. I was surprised when I found out that I actually have to type all this in the terminal. That is why I tried to be more explicit in the new readme.

stronk7:
Hi @iwasherefirst2,

nice improvement!

Would you mind, please, rebasing your commits on top of current master (I think you will face some minor conflicts to solve with that). And then proceed to squash current 6 commits, so everything can get the final review and, hopefully, land all-together?

Ciao :-)

stronk7:
Side comment about this patch, because for example it will affect some of my usual flows with moodle-docker.

Maybe the `.env` file should be setting the env variables **only if they aren't defined** ? I mean, basically act as "defaults" more that "environment" and only apply for them if undefined.

Real example: I've here some wrappers over moodle-docker that are basically a couple of nested loops changing the PHP versions (`MOODLE_DOCKER_PHP_VERSION `) and the databases (`MOODLE_DOCKER_DB`). I know that I can continue using it as is, as far as I don't create the `.env` file... everything will continue working the same.

But, if I normally want to work with `postgres` (for example) and I'm interested into the new file... then the wrapper above will stop working, because the variables in that file will, always, get precedence over the already specified ones.

So, just guessing if we should consider "renaming" all this from "env" to "defaults" and provide that slightly different behavior. For people not switching often it will behave 100% the same. But for crazy-switchers will make life easier.

Thoughts?

BTW, another little idea, is that maybe the `.env.example` template, better if it's not invisible and, also, maybe better if it has some comment about the possibility of setting up any other env variable there to make it explicitly clear.

Up to you, only a couple of details that came to my brain when reviewing current proposed patch.

Thanks for your contribution, much appreciated!

Ciao :-)

bin/create-moodle-config
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+if [ -f .env ]; 
+then
+    source .env
```
septatrix, line 5: This will not work as the variables are not exported (at least inside you sample env file)
Also applies to all following uses of `source .env`

README.md
```diff
@@ -18,27 +18,22 @@ This repository contains Docker configuration aimed at Moodle developers and tes
...
-export MOODLE_DOCKER_WWWROOT=/path/to/moodle/code
-# Choose a db server (Currently supported: pgsql, mariadb, mysql, mssql, oracle)
-export MOODLE_DOCKER_DB=pgsql
+1. Download or clone this repository
+2. Rename `.env.example` to `.env`
+3. Specify Moodle directory and db driver in `.env`
+4. Run `bin/create-moodle-config`. This copies the `config.php` into Moodle.
+You only have to call this command onces.
+4. Run `bin/moodle-docker-compose up -d` inside `moodle-docker` folder to create the container.
+5. You may have to call also `bin/moodle-docker-wait-for-db` if you use `oracle/mssql`.
```
septatrix, line 27: You may have to -> You should (at least to my knowledge you always want to do this if you execute these commands directly after each other)

README.md
```diff
@@ -18,27 +18,22 @@ This repository contains Docker configuration aimed at Moodle developers and tes
...
-```bash
-# Set up path to Moodle code
-export MOODLE_DOCKER_WWWROOT=/path/to/moodle/code
-# Choose a db server (Currently supported: pgsql, mariadb, mysql, mssql, oracle)
-export MOODLE_DOCKER_DB=pgsql
+1. Download or clone this repository
+2. Rename `.env.example` to `.env`
+3. Specify Moodle directory and db driver in `.env`
+4. Run `bin/create-moodle-config`. This copies the `config.php` into Moodle.
+You only have to call this command onces.
```
septatrix, line 25: This seems to me like a more complicated setup than what was here previously.

.env.example
```diff
@@ -0,0 +1,4 @@
+# Set up path to Moodle code
+MOODLE_DOCKER_WWWROOT=/path/to/moodle/code
+# Choose a db server (Currently supported: pgsql, mariadb, mysql, mssql, oracle)
+MOODLE_DOCKER_DB=pgsql
```
septatrix, line 4: These need an export before them if you want to source them from the scripts. dotenv files like you know them from php, python etc will not work with shells.

septatrix (changes requested):
I do not think this add any value (besides the addition to the gitignore file).
If people want to save these variables they are free to create a .env file themselves and source it beforehand. This is also what every seasoned dev probably already does. Therefore adding the entry to the gitignore file seems like a good idea and mentioning that this can be done in the docs will not hurt either. However adjusting all the scripts etc does not really add any value IMO.

skodak:
I have used a bit different approach in https://github.com/moodlehq/moodle-docker/pull/232

PR #135, opened 2020-09-02
Define some helpful config flags for running tasks.

paulholden:
They allow cron/tasks to be run immediately through the UI without additional configuration.


stronk7:
Merged, thanks @paulholden !

PR #151, opened 2021-01-17
Set valid default noreply address.

### LINKED ISSUE #121: On a newly installed instance, Front page summary and No-reply address are announced as new settings
Steps to reproduce:
* Set up a new Moodle-Docker instance
* Use the default config.php file from https://github.com/moodlehq/moodle-docker/blob/master/config.docker-template.php
* Initialize the instance for manual testing with the command from https://github.com/moodlehq/moodle-docker#use-containers-for-manual-testing
* Open the instance in the browser
* Login as admin
* Go to the notifications page

Expected result:
* You see the notifications page

Actual result:
* You are redirected to a "New settings page" which announces two new settings:
** $CFG->summary
** $CFG->noreplyaddress
image
* You can't just hit "save changes" as the default for $CFG->noreplyaddress is "noreply@localhost" which is not a valid email address. You have to change $CFG->noreplyaddress to a valid email address first

-----

Note:
This issue might also simply be a Moodle core issue and not a special Moodle-Docker issue. 
However, I report it here first as I have only encountered it in Moodle-Docker up to now and it affects especially Moodle-Docker admins who create instances on a daily basis.

If I will be told that this isn't a speciality of Moodle-Docker, I will create an issue in Moodle tracker for this behaviour. 

Thanks,
Alex

paulholden:
Fixes #121.

### DIFF
--- a/config.docker-template.php
+++ b/config.docker-template.php
@@ -31,6 +31,7 @@ $CFG->dataroot  = '/var/www/moodledata';
 $CFG->admin     = 'admin';
 $CFG->directorypermissions = 0777;
 $CFG->smtphosts = 'mailhog:1025';
+$CFG->noreplyaddress = 'noreply@example.com';
 
 // Debug options - possible to be controlled by flag in future..
 $CFG->debug = (E_ALL | E_STRICT); // DEBUG_DEVELOPER


stronk7:
Thanks, merged!

PR #153, opened 2021-03-23
bin: Correctly handle spaces in basedir and other refinements.

apteryks:
This also makes use of the Bash builtin extended test operator
features to streamline the conditionals.

* bin/moodle-docker-compose: Do not use literal '$' in the user messages.
(dockercompose): Rename variable to ...
(dockercompose_cmd): ... this, which is made an array, which eases
correctly handling quoting.
(filename): Quote.
* bin/moodle-docker-wait-for-app (basedir): Quote.  Chain the
conditions inside the same extended test.
* bin/moodle-docker-wait-for-db (basedir): Quote.  Do not use literal
'$' in the error message.

Signed-off-by: Maxim Cournoyer <maxim.cournoyer@savoirfairelinux.com>

### DIFF
--- a/bin/moodle-docker-compose
+++ b/bin/moodle-docker-compose
@@ -3,13 +3,13 @@ set -e
 
 if [ ! -d "$MOODLE_DOCKER_WWWROOT" ];
 then
-    echo 'Error: $MOODLE_DOCKER_WWWROOT is not set or not an existing directory'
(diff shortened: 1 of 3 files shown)


septatrix:
I also fixed this myself and wanted to open a PR before finding yours. It would be great to get this merged to avoid anyone else going through the same troubles

PR #156, opened 2021-05-03
Update app config to support Ionic 5

NoelDeMartin:
In the Apps Team, we've been working on [migrating the mobile app to ionic 5](https://tracker.moodle.org/browse/MOBILE-3320).

Whilst doing that, there are some things affecting this repository that have changed. For example, the [images on docker hub](https://hub.docker.com/r/moodlehq/moodleapp/tags?page=1&ordering=last_updated&name=ionic5) were previously bundling the source code and that's why it took them a while to get started. For ionic 5, we started bundling the compiled assets and the images are now a lot smaller and run a lot faster. As a side-effect of that, images are served on port 80. Other than that, some commands have changed slightly.

The mobile app with Ionic 5 hasn't been released yet, so I don't think this should be merged until that happens. But we can use this PR to discuss the approach and have everything ready for release day. That way, the day that the master branch in the mobile app repository becomes ionic 5, we can merge this and keep both repositories in sync.


stronk7:
Hi, just a side note about #161 , where we are planning to stop and remove Travis integration with this repository, now that we have it covered with GHA alternative.

NoelDeMartin:
Thanks @stronk7, I've updated the config to include GHA tests as well. I've also updated the strategy to detect the runtime (ionic 3 or ionic 5), and it will be automatic most of the time.

stronk7:
Merged, thanks, yay!

PS: Only thing I found a little bit confusing is the tags thing, having next, dev, next-dev, latest, test, next-test... but surely it's because I'm not used to the different alternatives and the target of each one. The docs page seems to explain it pretty well, so all right!

PR #158, opened 2021-05-08
Test php80 combinations with 311 and master

stronk7:
Move 311 to be handled as stable (no mater there is 1 week still before releasing v.3.11.0.

And then, add the php80 to all the highest combinations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,20 @@
-language: bash
+language: shell
+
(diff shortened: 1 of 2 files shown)


stronk7:
Self-merging, this PR is only updates to Travis to ensure php80 images are working ok, and I need this landed to continue advancing with other subproducts.

PR #159, opened 2021-05-17
Add GHA support to moodle-docker

stronk7:
Basically does the same than old Travis runs, with some differences:

- I've grouped the jobs into 3 categories (phpunit, behat, app).
- Each one has own setup/teardown/test shell script, instead of the
  previous "mammoth" setup and test scripts, that I've left unmodified
- Aiming to get the travis CI runs disabled because they are super slow
  (we only have 2 concurrent jobs allowed) and, while running, all the
  other moodlehq repos have to wait.
- I'll create an issue to remove travis support once this
  is accepted and working ok.
- Only change that I had to perform to existing stuff is raising the
  hardcoded sleep of 5 seconds to 10 seconds because I was getting some
  random failures with MySQL, needing more time to start.
- We should move to proper health-checks for database (#160) containers soon,
  much like app containers (#131).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,183 @@
+name: moodle-docker CI
+
+on: [push, pull_request]
+
+jobs:
+  PHPUnit:
+    runs-on: ubuntu-18.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # PostgreSQL (highest, lowest php supported)
+          - { branch: master,            php: "8.0", database: pgsql, suite: phpunit-full } # Full run only for mater.
+          - { branch: master,            php: "7.3", database: pgsql, suite: phpunit-full }
+          - { branch: MOODLE_311_STABLE, php: "8.0", database: pgsql, suite: phpunit } # Other branches, quicker run.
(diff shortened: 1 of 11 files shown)


stronk7:
You can see it passing @ https://github.com/stronk7/moodle-docker/actions/runs/849100015

And the very-last commit (36fe4b0) that makes this PR is running @ https://github.com/stronk7/moodle-docker/actions/runs/849335063

Ciao :-)

PS: Obviously... this PR only can run Travis tests, because GHA is not upstream, heh. ;-)

stronk7:
Have amended the commit to save setting up one variable... latest changes are being run now by Travis and https://github.com/stronk7/moodle-docker/actions/runs/849537749.

Ciao :-)

vmdef:
Thank you @stronk7. The structure is quite clear and the comment makes it even easier to understand the code. Only one minor typo that does not "[matter](https://github.com/moodlehq/moodle-docker/pull/159/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR14)".

stronk7:
> minor typo that does not "matter".

Whops, will try to fix it on next PR, thanks!

Issue #146, opened 2020-12-09
Xdebug 3.0.1 now fetched


djhipps:
The version of Xdebug now fetched from this Docker build will be 3.0.1. 
Could you please update the README to reflect something like the new config settings that will be required? 
There are mine for VSCODE on WSL2. As you can see they have dramatically changed and v2 settings will not work. (I am happy to do this if you want).

https://xdebug.org/updates
https://xdebug.org/docs/upgrade_guide

> ; Settings for Xdebug Docker configuration
> xdebug.mode=debug
> xdebug.start_with_request=yes
> xdebug.client_port=9003
> xdebug.remote_handler=dbgp
> xdebug.idekey=VSCODE
> xdebug.discover_client_host=false
> xdebug.client_host=host.docker.internal
> xdebug.log=/var/log/xdebug.log
> 
> {
>     "version": "0.2.0",
>     "configurations": [
> 
>         {
>             "name": "Listen for XDebug WSL",
>             "type": "php",
>             "request": "launch",
>             "port": 9003,
>             "log": true,
>             "externalConsole": false,
>             "pathMappings": {
>                 "/var/www/html": "${workspaceRoot}",
>             },
>             "hostname": "localhost"
>         },
>     ]
> }
> 
>

mattrice:
I can confirm this config works for XDebug 3.0.3 and VSCode 1.54.3

stronk7:
Hi @djhipps ,

this is being handled by #163, please take a look there in case there is anything  worth considering to be added there.

I'm closing this now, thanks!

adpe:
Thanks @djhipps 

Actually my PR will not set custom IDE configuration options, just the two needed options are set. Can you please try it out and check if it will work with VSCode? For my case I can confirm, that for PHPStorm no other configuration options must be set.

Greets
Adrian

PR #163, opened 2021-05-19
Update README.md due Xdebug 2 to 3 upgrade

adpe:
https://xdebug.org/docs/upgrade_guide

PR result of this message https://t.me/moodledev/90141

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,22 +215,17 @@ moodle-docker-compose exec webserver pecl install xdebug
 # Set some wise setting for live debugging - change this as needed
 read -r -d '' conf <<'EOF'
 ; Settings for Xdebug Docker configuration
-xdebug.coverage_enable = 0
-xdebug.default_enable = 0
-xdebug.cli_color = 2
-xdebug.file_link_format = phpstorm://open?%f:%l
-xdebug.idekey = PHPSTORM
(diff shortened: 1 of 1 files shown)


stronk7:
Just cross-linking to #146 because in that issue there were some extra config settings that we may want to have documented, feel free to discuss about anything here.

Ciao :-)

djhipps:
I have reviewed the above and confirm that 

> read -r -d '' conf <<'EOF'
> ; Settings for Xdebug Docker configuration
> xdebug.mode=debug
> xdebug.client_host=host.docker.internal
> xdebug.idekey=VSCODE
> EOF

is suffient for me and some of the settings in #146 were superflous.

stronk7:
So, the `idekey` is needed, correct? Or am I wrong and it's not needed?

Maybe it's the IDE the one requiring it? For example, I remember myself playing with it with `vim` and I think it was not needed (in fact I don't have it defined in my `php.ini` files.

What about adding it commented out, sort of:
```
; Some IDEs (list, of, ides, known, to, require, it...) may require to configure an idekey, uncomment if needed
; xdebug.idekey=MY_FAV_IDE_KEY
````

Ciao :-)

stronk7:
And merged, thanks!

Issue #166, opened 2021-06-16
Deprecated MOODLE_APP_VERSION variable not properly handled on Windows


NeillM:
If you use the MOODLE_APP_VERSION on windows the MOODLE_DOCKER_APP_VERSION variable is set with speech marks around it. i.e. as **"latest"**, rather than **latest** causing moodle-docker not to start with a message like:

ERROR: no such image: moodlehq/moodleapp:"latest": invalid reference format

NeillM:
I have submitted a pull request to fix this.

stronk7:
Related to this, I've created #168 to see if we can have Windows scripts covered in some way.

stronk7:
Closing, now that #167 has been merged, thanks!

PR #167, opened 2021-06-16
App fails to load on Windows using deprecated variable

NeillM:
Before this change if you set MOODLE_APP_VERSION to latest the MOODLE_DOCKER_APP_VERSION would be set to "latest" which means it failed to load the app image causing the docker images not to be loaded.

This is a fix for #166


stronk7:
Thanks!

Issue #170, opened 2021-06-23
Increase database engine versions / make tt


aspark21:
https://github.com/moodlehq/moodle-docker/blob/master/db.mariadb.yml#L8
https://github.com/moodlehq/moodle-docker/blob/master/db.mysql.yml#L8

stronk7:
Yeah, we have a similar request also for [moodle-ci-runner](https://github.com/moodlehq/moodle-ci-runner), that runs all our CI jobs, see [MDLSITE-6376](https://tracker.moodle.org/browse/MDLSITE-6376).

Surely both (they are different, but "parallel" enough) could be achieved in a similar way.

A) For some reason what I had been thinking was, more or less:

1. Create new, unique, `MOODLE_DOCKER_DB_VERSION` (say, 8, or 13, or 5.7.12...) variable.
2. If not set, each database will default to current values.
3. If set, then use it as tag for `MOODLE_DOCKER_DB` image (mysql, postgres...) to use.
4. Always look for `db.MOODLE_DOCKER_DB.MOODLE_DOCKER_DB_VERSION`  (db.mysql8.yml, db.oracle19.yml...) and use them if they exist.
5. Fallback to un-versioned `db.MOODLE_DOCKER_DB` (db.mysql.yml, db.oracle.yml...) if versioned doesn't exist.

I'm aware it's not the unique alternative. Maybe we could also:

B) Do everything with existing `MOODLE_DOCKER_DB`, making it to accept database:tag format.
C) Make a different `MOODLE_DOCKER_DB_VERSION` variable for each database.
D) ...

It would be really interesting to see if we can come with an agreement and get this implemented, yes!

Ciao :-)

aspark21:
Got a version of it working - https://github.com/aspark21/moodle-docker/blob/mysql8/db.mysql8.yml
Had to remove 2 settings deprecated in 8 and add one for auth - https://github.com/moodlehq/moodle-docker/compare/master...aspark21:100c005241743552d2fe2a08157557b5c5f4fb54

For that reason, we're going to need a separate image.

Just verifying it runs properly and I'll submit a MR with that first step.


Having a look at moodle-ci-runner, the version change would need to take place here:
https://github.com/moodlehq/moodle-ci-runner/blob/master/runner/master/run.sh#L283
https://github.com/moodlehq/moodle-ci-runner/blob/master/runner/master/run.sh#L299

The deprecated parameters were already removed to enable mariadb
[...]

PR #171, opened 2021-07-02
Make a MySQL 8 image available

aspark21:
Add a MySQL 8 image option

Part of #170


skodak:
this would be resolved also by https://github.com/moodlehq/moodle-docker/pull/225

PR #173, opened 2021-07-14
Make mariadb:10.5 sticky

stronk7:
New mariadb:10.6 makes the COMPRESSED row format read-only by default
because it's deprecated now and will be removed in mariadb:10.7

This is being tracked @ https://tracker.moodle.org/browse/MDL-72131

Once that issue is fixed and we switch to DYNAMIC or whichever the
final solution is, we'll unpin this.


sarjona (approved):
The fix looks good. Thanks, Eloy! :-*

PR #176, opened 2021-08-02
Use sticky selenium/standalone-chrome:3.141.59-20210713

stronk7:
New version3.141.59-2021729 is causing problems, so we need
to make it sticky to use a known to work version.


stronk7:
Self-merged, tests are now passing ok.

PR #183, opened 2021-09-01
Use NFS for Mac users

odeialba:
Docker is very slow for mac users (mainly when installing the db from scratch or setting up behat and phpunit and running them).
To fix that, we are able to use NFS. This PR uses NFS for Mac users instead of the old `:cached`.

Source: https://vivait.co.uk/labs/docker-for-mac-performance-using-nfs

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ export MOODLE_DOCKER_DB=pgsql
 # Ensure customized config.php for the Docker containers is in place
 cp config.docker-template.php $MOODLE_DOCKER_WWWROOT/config.php
 
+# [..] IMPORTANT: For Mac users see next point
+
 # Start up containers
 bin/moodle-docker-compose up -d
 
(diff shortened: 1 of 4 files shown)


crazyserver:
This is working very well for me, can we get that integrated?

mouneyrac:
The current Mac performance solution is to enable both the experimental settings (virtualisation and VirtioFS) in the preferences of Docker Desktop.

Docker Desktop Preference Settings

See https://www.docker.com/blog/speed-boost-achievement-unlocked-on-docker-desktop-4-6-for-mac/

skodak:
I find the performance on M2 mac satisfactory if "new virtualisation framework" and "VirtioFS" enabled from experimental settings.

crazyserver:
Thanks @mouneyrac  and @skodak , 
 I also think this issue can be closed.

stronk7:
Thanks all!

Issue #154, opened 2021-04-27
Update version of the Firefox profile to make use of newer webdriver


aspark21:
https://moodle.org/mod/forum/discuss.php?d=417286 states "We will be updating the Moodle-docker toolkit over the coming days too."

I assume the intent was to update the sample config.php to make use of https://github.com/andrewnicols/moodle-browser-config

andrewnicols:
Hi @aspark21 ,

The change mentioned there is actually to switch to a different version of Firefox.

Older versions of Firefox are no longer supported with the newer webdriver so we need to modify the Firefox profile at https://github.com/moodlehq/moodle-docker/blob/master/base.yml#L26 to use a different image.

We do also need to consider moving this one to a different yml file so that anyone testing against legacy versions of Moodle can still use Firefox.

aspark21:
Hi @andrewnicols 

Ah cool. 

I see moodle-plugin-ci uses 3.141.59 https://github.com/moodlehq/moodle-plugin-ci/blob/master/src/Command/BehatCommand.php#L41

So I'm hoping this Fatal Error should be resolved if we switch firefox versions - https://gitlab.com/jezhops/moodle-theme_adaptable/-/issues/159#note_561876589 ?

I've forked this and increased the version quickly - https://github.com/aspark21/moodle-docker/commit/b5f4b0ad4935eb87376fea3225a2f777c692b011

And I've set off the job to run again with that. 🤞 that allows it to complete this evening

aspark21:
Updated the title of the story to better reflect the change needed.

Unfortunately the simple hotfix did not work.

We get these errors under Boost all the time now, random how long the behat runs get to go on for but always end up hitting this.

```
 Running 2 parallel behat sites:
    [behatrun1] vendor/bin/behat --config /var/www/behatdata/behatrun1/behat/behat.yml --format="pretty" --out="report_behat_1.txt" --format="junit" --out="junit_behat_1"
    [behatrun2] vendor/bin/behat --config /var/www/behatdata/behatrun2/behat/behat.yml --format="pretty" --out="report_behat_2.txt" --format="junit" --out="junit_behat_2"
    Fatal error: Uncaught Facebook\WebDriver\Exception\WebDriverCurlException: Curl error thrown for http DELETE to /session/a172997a-95ea-4162-a3da-fbdaac17da65
[...]

PR #194, opened 2021-12-02
Update Firefox

aspark21:
Quick first step for #154 is to update the default Firefox version.

Will submit the rewiring separately


stronk7:
My understanding is that we are not going to need this once #195 is ready (and supersedes this), correct?

aspark21:
Correct.

This was purely the quick fix which would allow tests on the master branch to pass, as master currently fails e.g. https://github.com/aspark21/moodle-docker/actions/runs/1394569767

Which make it seem new contributions break things when in fact master is failing as is. 

But #195 is the more long term improvement & more flexible way of dealing with that. Wasn't sure how long #195 would take to implement so I put this one in as a fallback.

aspark21:
Closing as #195 is ready

Issue #191, opened 2021-11-16
Javascript behat tests failing on newer versions of Moodle


NeillM:
All Javascript based behat tests seem to be failing if running a version of Moodle that contain https://tracker.moodle.org/browse/MDL-65943

**Steps to reproduce**

1. Checkout MOODLE_310_STABLE
2. Start moodle-docker using the default settings (pointing at it)
3. Initialise behat
4. Run `bin/moodle-docker-compose exec webserver vendor/bin/behat --config /var/www/behatdata/behatrun/behat/behat.yml --tags="@javascript&&@mod_assign"`

All the tests will fail

![image](https://user-images.githubusercontent.com/1038050/142013437-9d321fc0-9417-415b-afdf-411c7dfc07d7.png)

with a message similar to:

`BeforeStep # behat_hooks::before_step()
      Javascript code and/or AJAX requests are not ready after 10 seconds. There is a Javascript error or the code is extremely slow (theme_boost/loader:theme_boost/drawer). If you are using a slow machine, consider setting $CFG->behat_increasetimeout. (Exception)`

If you checkout commit cede428cc4b9d339d81f35025f00522a2075cce4 which is the one immediately before the merge of MDL-65943 (and conveniently the previous weekly release) the same tests will pass.

It seems to be somewhat environmental since this does not appear to be failing when I run the tests outside of moodle-docker

NeillM:
Thanks Alistair I'll give that patch a try tomorrow

NeillM:
The tests do run normally with Alistair's patch applied. I'm guessing Moodle got made incompatible with the version of Firefox that is used by default.

PR #195, opened 2021-12-02
Selenium versions rebased

aspark21:
Rebased #60 and made Readme docs more explicit that the tag is the selenium version not the browser version. (hoping this addresses https://github.com/moodlehq/moodle-docker/pull/60#issuecomment-331009133)

Been discussing in #154 

I will go off and test it out against Windows/Mac/Linux but would be useful to know if any further tweaks are required.


aspark21:
Could rename MOODLE_DOCKER_BROWSER_TAG to MOODLE_DOCKER_SELENIUM_TAG but since the MOODLE_DOCKER_BROWSER_TAG is extracted from MOODLE_DOCKER_BROWSER there is a logic to that name.

stronk7:
Awesome sauce @aspark21 !

Only details that comes to my brain is that, recently, SeleniumHQ has started to stop building the 3.x images and they are only building the 4.x ones.

But... we aren't ready yet to 4.x ones (there are some ongoing PRs to get them working by default).

So great that you've pinned the defaults for both to be 3.x. In fact, unless I'm wrong, I think you can make both them to be "3" and done. And, maybe, those defaults should be explained in the README, clearly stating that the default is "firefox:3".

But other than that, if compose works ok for you for both unix and windows... yayayay!

stronk7:
For reference, this will close: #154, #191, #194, #20 and #60

aspark21:
Linux full run has completed.
Ran it with `MOODLE_DOCKER_BROWSER: firefox:3`
That seemed to have **reduced our Behat failures** by 4 but full build (master +120 plugins) run increased runtime from 9h57m to 11h23m. We probably have some night time system processes slowing things down so won't take note of that.
Either way, this means we can indeed unpin the selenium version for Firefox. 

Have changed that accordingly, including the README.

Moving onto Windows testing

aspark21:
Windows testing complete.
behat2

behat

Should be all clear to merge.

aspark21:
Happy New Year!

Giving this a nudge as this is ready to be integrated. We've done a good 100 runs with this while reworking our pipelines and this has been working well.

stronk7:
Happy new year!

And thanks for all the work here, merged!

I'm going to close related issues (#154, #191, #194, #20 and #60).

Ciao :-)

Issue #84, opened 2018-05-03
Enable parallel behat runs


marxjohnson:
I tried the following:
```
export MOODLE_DOCKER_WWWROOT=/path/to/moodle/code
export MOODLE_DOCKER_DB=pgsql
export MOODLE_DOCKER_BROWSER=chrome
cp config.docker-template.php $MOODLE_DOCKER_WWWROOT/config.php
bin/moodle-docker-compose up -d
bin/moodle-docker-wait-for-db
bin/moodle-docker-compose run webserver php admin/tool/behat/cli/init.php --parallel=4
bin/moodle-docker-compose run webserver php admin/tool/behat/cli/run.php
```
I was hoping this would launch a 4-threaded behat run.  The init step ran fine and installed the 4 databases, but the run step just resulted in this message:
```
Running single behat site:


  [Behat\Testwork\ServiceContainer\Exception\ConfigurationLoadingException]
  The requested config file does not exist


To re-run failed processes, you can use following command:
php admin/tool/behat/cli/run.php --rerun
```

I'm guessing there's a need to spin up extra Selenium containers and monkey around with the config.php in order to get this working.  It would be lovely if there was a command to do this!

aspark21:
It's been a while but this does currently already work. 
Have been routinely using 2 parallel runs for the last 18 months and just completed testing 6-10 parallel runs. 

I was thinking the only other thing that could be useful to implement for this was adding support for a timing file so did quite a few tests on our runners (12CPUs 16GBs RAM)

| Parallel runs      | without  | with timing file | Moodle version |
| -----------        | -------- | --------             | --------             |
| 2                        | 11h31m   |  NA                   | master/4.0dev   |
| 2                        | 9h31m    |  NA                   | 3.9                      |
| 6                        | 4h16m    |  NA                   | master/4.0dev   |
| 6                        | 3h27m    |  NA                   | 3.9                      |
| 7                        | 3h59m   |  3h15m              | master/4.0dev   |
| 8                        | 3h20m   |  2h57m             | master/4.0dev   |
| 9                        | 3h44m    |  NA                   | master/4.0dev   |
| 10                       | 2h59m    |  **2h33m**            | master/4.0dev   |
| 12                       | 2h57m    |  2h28m            | master/4.0dev   |

(Will take the 10 parallel with timing file into service)

There is some benefit to having the timing file (30mins or so) but I'm not sure it would be that beneficial to have anything specifically setup in moodle-docker for it.

[...]

PR #196, opened 2021-12-06
Faildump mapping

aspark21:
I know there's the web browser approach to loading the faildumps but we're using moodle-docker within our CI pipeline so looking to retrieve the files in bulk without opening a browser. Just another option. Found an old commit from Andrew that had started this and updated this to align with the path which has since been defined for faildumps.

Tested in our pipeline and works well.

Github Actions will have the usual failures because #195 not yet in but this passes here - https://github.com/aspark21/moodle-docker/commits/merge-them-all / https://github.com/aspark21/moodle-docker/actions/runs/1535306113 - where I've merged 195, this and the mysql8 change(171) into one branch.


bin/moodle-docker-compose
```diff
@@ -128,6 +128,11 @@ then
     dockercompose="${dockercompose} -f ${basedir}/phpunit-external-services.yml"
 fi
 
+if [[ ! -z "$MOODLE_DOCKER_BEHAT_FAILDUMP" ]];
```
scara, line 131: Hi @aspark21,
what about testing even if it is actually a directory too?
Kind of `[[ -d "$MOODLE_DOCKER_BEHAT_FAILDUMP" ]]`.

HTH,
Matteo

bin/moodle-docker-compose.cmd
```diff
@@ -64,6 +64,10 @@ IF NOT "%MOODLE_DOCKER_PHPUNIT_EXTERNAL_SERVICES%"=="" (
     SET DOCKERCOMPOSE=%DOCKERCOMPOSE% -f "%BASEDIR%\phpunit-external-services.yml"
 )
 
+IF NOT "%MOODLE_DOCKER_BEHAT_FAILDUMP%"=="" (
```
scara, line 67: Hi @aspark21,
same as above, what about adding a dir check even here?
Kind of `exist "%MOODLE_DOCKER_BEHAT_FAILDUMP%\nul"`.

HTH,
Matteo

bin/moodle-docker-compose
aspark21: Fair point updated accordingly

bin/moodle-docker-compose.cmd
aspark21: Aligned the checks with those done for MOODLE_DOCKER_WWWROOT but nested as this is not a mandatory env variable

aspark21:
re-based on master now that #195 integrated so all tests are passing 😄 

 I believe I've addressed all of your comments @scara 

This has dual benefit of allowing access to faildumps but also to use a timing file when running test suites - https://github.com/moodlehq/moodle-docker/issues/84#issuecomment-991993273 (that might warrant some documentation / tweak to config.docker-template.php in a separate pull request)

stronk7:
Here we go, thanks @aspark21 ! 😎

PR #203, opened 2022-02-07
Use Compose v2 if it is available in the system.

### LINKED ISSUE #202: Transition to use Compose v2 in moodle-docker-compose wrapper
New Go-based compose implementation has become a default branch and primary development focus (https://docs.docker.com/compose/cli-command/):
> * New features and bug fixes will only be considered in the Compose V2 code base.
> * Docker Compose V2 will be the default setting in Docker Desktop for Mac and Windows. You can still opt out through the Docker Desktop UI and the CLI. This means, when you run docker-compose, you will actually be runningdocker compose.
> * Compose V2 will be included with the latest version of the Docker CLI. You can use [Compose Switch](https://docs.docker.com/compose/cli-command/#compose-switch) to redirect docker-compose to docker compose.
> * [Compose V2 branch](https://github.com/docker/compose/tree/v2) will become the default branch.
> * Docker Compose V1 will be maintained to address any security issues.

Looks like `docker-compose` alias will exist for some time (see [End of Life roadmap issue](https://github.com/docker/roadmap/issues/257)), however it is now a translating script for `docker compose` rather than a separate tool and [requires a separate installation](https://docs.docker.com/compose/cli-command/#compose-switch). Compose v2 (`docker compose`) will be a part of Docker CLI by default. For Linux it still needs to be [installed as plugin](https://docs.docker.com/compose/cli-command/#install-on-linux) in a meantime, but for Win and Mac it is [already included](https://docs.docker.com/compose/cli-command/#install-on-mac-and-windows).

**What this means for moodle-docker**

There is a possiblity that some users may not be able to execute `docker-compose` at some point which `bin/moodle-docker-compose` rely upon, unless they have [Compose Switch](https://docs.docker.com/compose/cli-command/#compose-switch) installed. Therefore, I suggest to address the issue earlier by:
1. Use `docker compose` v2 command by default and fallback to use `docker-compose` if former one is not available in the system.
2. Amend docs explaining above in requirements.

kabalin:
No changes to windows script as they seem maintaining alias which is using 'docker compose' insternally, given there is an option to disable v2 behavior by calling `docker-compose disable-v2` (https://docs.docker.com/compose/cli-command/#install-on-mac-and-windows), so for now it is fine to use it as is.

Fixes #202

### DIFF
--- a/bin/moodle-docker-compose
+++ b/bin/moodle-docker-compose
@@ -24,8 +24,15 @@ done
 basedir="$( cd -P "$( dirname "$SOURCE" )/../" && pwd )"
 export ASSETDIR="${basedir}/assets"
 
-
-dockercompose="docker-compose -f ${basedir}/base.yml"
+# Test if we have docker compose v2, and keep quiet if we don't.
+ver=$(docker compose version > /dev/null 2>&1 && docker compose version --short) || true
+if [[ $ver =~ ^v2 ]]; then
+  dockercompose="docker compose"
+else
+  echo 'Compose v2 is not available in Docker CLI, falling back to use docker-compose script'
+  dockercompose="docker-compose"
+fi
+dockercompose="${dockercompose} -f ${basedir}/base.yml"
 dockercompose="${dockercompose} -f ${basedir}/service.mail.yml"
 
 # PHP Version.
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ This repository contains Docker configuration aimed at Moodle developers and tes
 * Backed by [automated tests](https://travis-ci.com/moodlehq/moodle-docker/branches)
 
 ## Prerequisites
-* [Docker](https://docs.docker.com) and [Docker Compose](https://docs.docker.com/compose/) installed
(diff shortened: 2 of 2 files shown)


bin/moodle-docker-compose
```diff
@@ -25,7 +25,14 @@ basedir="$( cd -P "$( dirname "$SOURCE" )/../" && pwd )"
 export ASSETDIR="${basedir}/assets"
 
 
-dockercompose="docker-compose -f ${basedir}/base.yml"
+docker compose version 2>&1 >/dev/null
```
dahrens, line 28: This does not work as intended, because of line 2: `set -e`. For `help set`:

> -e  Exit immediately if a command exits with a non-zero status.

kabalin: Thanks @dahrens, very good point. I made necessary changes.

To validate new approach I used small script which I tested on the machine with v2 `docker compose` and without:
```
#!/bin/bash
set -e

ver=$(docker compose version > /dev/null 2>&1 && docker compose version --short) || true

if [[ $ver =~ ^v2 ]]; then
  dockercompose="docker compose"
else
  echo 'Compose v2 is not available in Docker CLI, falling back to use docker-compose script'
  dockercompose="docker-compose"
fi

echo $dockercompose
```

This gave output matching the presence of docker compose v2.

stronk7:
Merged, thanks all!

PR #211, opened 2022-05-04
Correct Github URLs to always specify https protocol.

paulholden:


stronk7:
Thanks, @paulholden !

PR #212, opened 2022-05-24
Add 400_STABLE to CI tests

stronk7:
(thanks to @NoelDeMartin for updating the app ones  - and docs - to current usage)


stronk7:
Self-merged, we are green now, yay!

PR #216, opened 2022-06-08
Update Moodle Docker Compose V2 Check

### LINKED ISSUE #215: Compose Version Detection WSL Docker Desktop
Hi Team,

Each time I run `moodle-docker-compose` I am prompted with "Compose v2 is not available in Docker CLI, falling back to use docker-compose script".
This appears to be due to the check at [moodle-docker-compose#L29](https://github.com/moodlehq/moodle-docker/blob/master/bin/moodle-docker-compose#L29) checking that the version starts with `v2` whereas my result from `docker compose version --short` is `2.6.0` with no leading v. I'm not sure if this is the case for a standalone Docker installation as I'm running this on WSL2 (Ubuntu 20.04.4) under windows 10 with Docker Desktop providing Docker via their WSL Integration.

NickLambellGCH:
This makes the `v` prefix optional in checks against the return value of `docker compose version --short` as it was removed in recent versions. e.g. The expected output was `v2.2.3` but is now simply `2.6.0`.

Closes #215.

### DIFF
--- a/bin/moodle-docker-compose
+++ b/bin/moodle-docker-compose
@@ -26,7 +26,7 @@ export ASSETDIR="${basedir}/assets"
 
 # Test if we have docker compose v2, and keep quiet if we don't.
 ver=$(docker compose version > /dev/null 2>&1 && docker compose version --short) || true
-if [[ $ver =~ ^v2 ]]; then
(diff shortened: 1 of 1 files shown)


stronk7:
Thanks @NickLambellGCH !

PR #220, opened 2022-07-24
Update to Oracle 21c

aspark21:
Fixes moodlehq#219

### DIFF
--- a/db.oracle.yml
+++ b/db.oracle.yml
@@ -5,4 +5,4 @@ services:
       MOODLE_DOCKER_DBTYPE: oci
       MOODLE_DOCKER_DBNAME: XE
   db:
-    image: moodlehq/moodle-db-oracle-r2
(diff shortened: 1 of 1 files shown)


stronk7:
Just noting that the 21c image is really slower (>25%) than then 11g one and that makes some behat tests to fail often (timeout, things not ready in time...).

Some attempts to improve its performance are still being done, but not luck, so far (specifically, it seems that access to dictionary tables, that we use intensively in tests, resets... are 4x slower).

Also, I've seen some strange failures happening in the 39_STABLE branch with it. Haven't had time to investigate.

Just noting it... those notes don't affect this PR. Only potential idea is if we should try to use it only for 4.1dev (master) and up, keeping the old one for previous branches.... but that's something that we can do later.

Ciao :-)

stronk7:
Merged, thanks!

Issue #160, opened 2021-05-17
Implement proper health-check for databases


stronk7:
Similar to #131 , but to ensure databases are available.

PR #221, opened 2022-07-25
Verify oracle availability using "DATABASE IS READY TO USE!"

stronk7:
We have ensured that all the oracle database images output
at the end of the init script the "DATABASE IS READY TO USE!"
text. That has been done by:

https://github.com/moodlehq/moodle-db-oracle/pull/13

(Because Oracle 11g was missing that string)

So, now that the string is available in all moodle-db-oracle
images we can start using it here, in moodle-docker.

Still note this is not the ideal wait/health-check system
and we'll have to move to a proper one. See:

https://github.com/moodlehq/moodle-docker/issues/160

But it's better than nothing, and the current one based
on the "Database opened." string was not working ok for
Oracle 21c (that requires 2 startups).

### DIFF
--- a/bin/moodle-docker-wait-for-db
+++ b/bin/moodle-docker-wait-for-db
@@ -14,7 +14,7 @@ then
     $basedir/bin/moodle-docker-compose exec -T db /wait-for-mssql-to-come-up.sh
 elif [ "$MOODLE_DOCKER_DB" = "oracle" ];
 then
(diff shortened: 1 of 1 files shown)


stronk7:
If automated tests for the master branch with Oracle now pass... it means the detection of the new string is working as expected.

Before this PR those jobs were failing 50% of time, assuming that the database was ready too early.

Ciao :-)

PR #231, opened 2022-10-07
simplify basedir variable setting

skodak:
I had trouble understanding the magic behind current basedir logic, so I invented a simpler version. It expects that only the executables from bin/* are symlinked or the whole moodle-docker. Symlinking just the bin directory does not make sense to me, they should just add the bin/ to their search path most likely instead.

I will be be using basedir it future patches, the current one seems to be a bit too long to copy/paste into all those other scripts such as waiting for db and app.

I can confirm it works fine in macOS and linux.

### DIFF
--- a/bin/moodle-docker-compose
+++ b/bin/moodle-docker-compose
@@ -1,6 +1,12 @@
 #!/usr/bin/env bash
 set -e
 
+# First find out if this was called from symlink,
+# then find the real path of parent directory.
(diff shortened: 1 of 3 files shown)


stronk7:
Merged, thanks!

skodak:
hehe, the whole point of improving moodle-docker is so that I do not have to install homebrew or macports in my new M2 mac

PR #232, opened 2022-10-07
Add support for ./moodle-docker.env

skodak:
See the README.md for more information how to use it.

My main objectives were:

- do not require setting MOODLE_DOCKER_WWWROOT when executing compose from Moodle checkout dir
- do not require COMPOSE_PROJECT_NAME when having multiple Moodle checkouts dir
- allow the db default to be stored in some file in Moodle checkout dir

NOTE: this feature is not supported in Windows because  I do not know how to parse the env file there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,36 @@ to the instance you expect to. See
 [envvars](https://docs.docker.com/compose/reference/envvars/)
 to see more about `docker-compose` environment variables.
 
+## Environment setting defaults via ./moodle-docker.env
+
+In Linux and macOS it is also possible to created `moodle-docker.env` file
+in directory where `bin/moodle-docker-compose` is executed, the file contents are
+then used as default environment values.
+
+If `moodle-docker.env` is in Moodle code root, then MOODLE_DOCKER_WWWROOT default
+is set to Moodle directory and COMPOSE_PROJECT_NAME defaults to Moodle directory name.
+
+Example:
+
+1. Create `/path/to/moodle/moodle-docker.env` file with the following content:
+```
+# file location /path/to/moodle/moodle-docker.env
+MOODLE_DOCKER_DB=pgsql
+```
+2. Copy config.docker-template.php file to `/path/to/moodle/config.php`
+3. Start containers:
+```bash
+cd /path/to/moodle
+/path/to/moodle-docker/bin/moodle-docker-compose up -d
+```
+4. Switch to MariaDB temporarily:
+```bash
+cd /path/to/moodle
+/path/to/moodle-docker/bin/moodle-docker-compose down
+export MOODLE_DOCKER_DB=mariadb
+/path/to/moodle-docker/bin/moodle-docker-compose up -d
+```
+
 ## Use containers for running behat tests
 
 ```bash
--- a/bin/include/env.cmd
+++ b/bin/include/env.cmd
@@ -1,0 +1,28 @@
+@ECHO OFF
+
+IF EXIST "moodle-docker.env" (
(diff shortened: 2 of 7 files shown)


jpahullo:
Hi @skodak,

According to docker-compose documentation, .env are also loaded in a prioritized way of loading variables. This is why I suggested add support for .env on #80. Natively it would be supported by Windows platforms.

However, it is a big and final step towards supporting several Moodle versions locally. Super! Congrats!

skodak:
hi @jpahullo , docker .env loading is not applicable in modole-docker-compose script because we need the environment variables BEFORE the real docker-compose is executed to create configuration for it.

With my patch you can still use .env, but the moodle-docker-compose script cannot use values from it.

I think I am onto something with the Windows support, I might try to make it work this weekend unless somebody volunteers.

jpahullo:
Hi @skodak ! You're totally right! I missed this part when commenting.

Thanks!!!

skodak:
I should have Windows support for this issue later today

skodak:
argh, not there yet, I ned to find out how to check if %%1 env variable is set

stronk7:
(sorry for the delay, @skodak, will be looking to this ASAP!)

jpahullo:
Any news from this?

Thanks a lot for your work!

andrewnicols:
See also #249 which uses shdotenv to get a better support for dotenv files.

stronk7:
Hi,

is there any reason we cannot make the env file to be a truly "dotenv" (`.env`) one?

Also is the "define only if not set" fallback respected?

1. Env variables already defined via export/set.
2. Env files support introduced by this issue.
3. moodle-docker own defaults.

Note I'm also looking to #249, really both are similar (add to the environment some variables, if they are not defined), using 2 different approaches.

I like here that it supports (not tested) Windows and that there aren't external dependencies. In the other side, surely the supported variables will be a little more limited, but I also think that it will be enough for what moodle-docker needs (simple 1-word contents).

Ciao :-)

jpahullo:
> Hi,
> 
> is there any reason we cannot make the env file to be a truly "dotenv" (`.env`) one?
> 
> Also is the "define only if not set" fallback respected?
> 
> 1. Env variables already defined via export/set.
> 2. Env files support introduced by this issue.
> 3. moodle-docker own defaults.
> 
> Note I'm also looking to #249, really both are similar (add to the environment some variables, if they are not defined), using 2 different approaches.
> 
> I like here that it supports (not tested) Windows and that there aren't external dependencies. In the other side, surely the supported variables will be a little more limited, but I also think that it will be enough for what moodle-docker needs (simple 1-word contents).
> 
> Ciao :-)

+1

mattporritt:
Hi @skodak ,
Thanks for working on this. I’ve given it a review. In addition to @stronk7 comment about confirming the fallback precedence:
The main change required here is to include the code in `include/env.sh` into the `moodle-docker-compose`. Including it via source is an unnecessary abstraction in this case.

Then (taking a couple of pointers from #249):

- Make an `.env.example` file and commit that
- Update the readme to refer to using and renaming the `.env.example` file
- Update the script that parses the file to expect `.env` as the file name
- Add `.env` to .`gitignore` so no one accidentally ever raises a PR with their credentials in it.

Cheers,
Matt P

skodak:
sorry @mattporritt , but the `include/env.sh` is essential when you need to use same environment in another shell script, such as waiting for moodle app

in any case I have decided to maintain my own form because I need this for all my development, so I will not be submitting any more pull requests here

PR #233, opened 2022-10-08
XDebug client_port

meirzamoodle:
In order to use Xdebug, we need to disable the firewall or open the port used in the xdebug.client_port.

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,6 +265,8 @@ read -r -d '' conf <<'EOF'
 ; Settings for Xdebug Docker configuration
 xdebug.mode = debug
 xdebug.client_host = host.docker.internal
+; Some IDEs (eg PHPSTORM, VSCODE) may require configuring an IDE key, uncomment if needed
+; xdebug.idekey=MY_FAV_IDE_KEY
 EOF
 moodle-docker-compose exec webserver bash -c "echo '$conf' >> /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini"
 
--- a/README.md
+++ b/README.md
@@ -273,7 +273,13 @@ moodle-docker-compose exec webserver docker-php-ext-enable xdebug
 moodle-docker-compose restart webserver
 ```
 
-While setting these XDebug settings depending on your local need, please take special care of the value of `xdebug.client_host` which is needed to connect from the container to the host. The given value `host.docker.internal` is a special DNS name for this purpose within Docker for Windows and Docker for Mac. If you are running on another Docker environment, you might want to try the value `localhost` instead or even set the hostname/IP of the host directly.
+While setting these XDebug settings depending on your local need, please take special care of the value of `xdebug.client_host` which is needed to connect from the container to the host. The given value `host.docker.internal` is a special DNS name for this purpose within Docker for Windows and Docker for Mac. If you are running on another Docker environment, you might want to try the value `localhost` instead or even set the hostname/IP of the host directly. Please turn off the firewall or open the port used in the `xdebug.client_port`.
+
+Open the port (9003 is the default one) by using the example command for Linux Ubuntu:
+```
+sudo ufw allow 9003
+```
+
 
 After these commands, XDebug ist enabled and ready to be used in the webserver container.
 If you want to disable and re-enable XDebug during the lifetime of the webserver container, you can achieve this with these additional commands:


scara:
Hi @meirzamoodle,
did you add the `client_port` setting for documentation purposes e.g. firewall settings?
`9003` is the default port.

TIA,
Matteo

meirzamoodle:
Hi @scara,

Yes, the changes for example how to open the port.

meirzamoodle:
Actually, there are other things to do if we're working with VSCode, Need to add a configuration in launch.json (configuration file from PHP debug extension), but I'm not sure if should i add it in the README file.

README.md
```diff
@@ -232,6 +232,7 @@ read -r -d '' conf <<'EOF'
 ; Settings for Xdebug Docker configuration
 xdebug.mode = debug
 xdebug.client_host = host.docker.internal
+xdebug.client_port = 9003
```
stronk7, line 235: 1. I would remove this from here (let's not add default values).

README.md
```diff
@@ -240,7 +241,13 @@ moodle-docker-compose exec webserver docker-php-ext-enable xdebug
 moodle-docker-compose restart webserver
 ```
 
-While setting these XDebug settings depending on your local need, please take special care of the value of `xdebug.client_host` which is needed to connect from the container to the host. The given value `host.docker.internal` is a special DNS name for this purpose within Docker for Windows and Docker for Mac. If you are running on another Docker environment, you might want to try the value `localhost` instead or even set the hostname/IP of the host directly.
+While setting these XDebug settings depending on your local need, please take special care of the value of `xdebug.client_host` which is needed to connect from the container to the host. The given value `host.docker.internal` is a special DNS name for this purpose within Docker for Windows and Docker for Mac. If you are running on another Docker environment, you might want to try the value `localhost` instead or even set the hostname/IP of the host directly. Please turn off the firewall or open the port used in the xdebug.client_port.
+
+Open the port by using the example command for Linux Ubuntu:
```
stronk7, line 246: And, maybe would add here:

`Open the port (9003 is the default one) by using the example command for Linux Ubuntu:`

stronk7 (changes requested):
Hi @meirzamoodle,

I must confess that I'm not 100% sure if this sort of information corresponds here, this can become a thick book if every single detail is added, not to talk about individual IDEs/environment differences.

In any case, ok, let's give to this a pass. If things become out of control, we always can apply scissors in the future.

Please take a look to the suggestions and this will be ready to be merged. Thanks for your collaboration!

meirzamoodle:
> Hi @meirzamoodle,
> 
> I must confess that I'm not 100% sure if this sort of information corresponds here, this can become a thick book if every single detail is added, not to talk about individual IDEs/environment differences.
> 
> In any case, ok, let's give to this a pass. If things become out of control, we always can apply scissors in the future.
> 
> Please take a look to the suggestions and this will be ready to be merged. Thanks for your collaboration!

Hi @stronk7. 

I just updated the changes based on your suggestion. Cheers!

README.md
```diff
@@ -240,7 +240,13 @@ moodle-docker-compose exec webserver docker-php-ext-enable xdebug
 moodle-docker-compose restart webserver
 ```
 
-While setting these XDebug settings depending on your local need, please take special care of the value of `xdebug.client_host` which is needed to connect from the container to the host. The given value `host.docker.internal` is a special DNS name for this purpose within Docker for Windows and Docker for Mac. If you are running on another Docker environment, you might want to try the value `localhost` instead or even set the hostname/IP of the host directly.
+While setting these XDebug settings depending on your local need, please take special care of the value of `xdebug.client_host` which is needed to connect from the container to the host. The given value `host.docker.internal` is a special DNS name for this purpose within Docker for Windows and Docker for Mac. If you are running on another Docker environment, you might want to try the value `localhost` instead or even set the hostname/IP of the host directly. Please turn off the firewall or open the port used in the xdebug.client_port.
```
scara, line 243: Suggestion: add the backticks (`) around `xdebug.client_port`.

HTH,
Matteo

meirzamoodle: Hi @scara . I'm sorry for the late response. I had wrapped `xdebug.client_port` with the backticks.

NeillM:
I have manged to get remote debugging working on PHPStorm with only a single change to the documented steps in the current README, that was to add an IDE key to the configuration, i.e.:

`xdebug.idekey = moodle`

However that might not be technically needed, but is probably useful in my case where I can have several different docker servers running with xdebug enabled.

My current setup has everything running on WSL2 containers.

stronk7:
I remember the `idekey` detail was discussed somewhere (relatively recently). Found it @ https://github.com/moodlehq/moodle-docker/pull/163 ... maybe we can go and add a side comment here about it (for whichever integration needs it).

NeillM:
The idekey may well not be needed (I didn't try without it) at least in PHPStorm it looks as though it should not be required

 I suspect you might want one in cases where you are running for example multiple moodle-docker instances for different versions of Moodle or different patches so that only debugging calls for the code you are looking at get to the correct IDE (in this sort of case you would probably want different keys for each docker instance)

Or if you have completely different different applications running on various docker instances (for example I have a Moodle and our online exams system)

stronk7:
I'm happy if we add it (commented) with something saying that, depending on the IDE or environments, it may be needed to setup it.

Would you, @meirzamoodle ? I think that with that little addendum... this can finally land... ciao :-)

meirzamoodle:
Hi, @stronk7. 

I already planned to complete this last week. Sorry for the late git push. Cheers.

stronk7:
Thanks @meirzamoodle,

it looks perfect, I'll land this as soon as GHA tests end.

Ciao :-)

stronk7 (approved):

PR #234, opened 2022-10-13
Bump to actions/checkout@v3

stronk7:
Related to this deprecation of NodeJS 12 for GHA actions:

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

While it doesn't affect to this repository, let's bump to the new v3 version now so we keep it updated.


stronk7:
Self merged, trivial.

PR #241, opened 2022-12-02
Include the BBB Mock container

andrewnicols:


config.docker-template.php
```diff
@@ -97,4 +97,13 @@
     define('TEST_ENROL_LDAP_DOMAIN', 'ou=Users,dc=openstack,dc=org');
 }
 
+if (getenv('BBBMOCK')) {
```
stronk7, line 100:
```suggestion
if (getenv('MOODLE_DOCKER_BBB_MOCK')) {
```

I'd suggest to use the very same name and then add it to the `bbb-mock.yml`file. See above.

bbb-mock.yml
```diff
@@ -0,0 +1,4 @@
+version: "2"
+services:
+  bbbmock:
```
stronk7, line 3:
```suggestion
  webserver:
    environment:
      MOODLE_DOCKER_BBB_MOCK: "true"
  bbbmock:
```

Let's declare the env variable here, so `config.php` is able to get it.

stronk7 (changes requested):
Hi,

it looks nearly ok. But the `BBBMOCK` used in the `config.php` template does not exist.

Because of that, and following a similar path used for other env variables, I've suggested to create the env variable (with the same name) in the new yml file and then just read it.

config.docker-template.php
andrewnicols: Agree - thanks.

mattporritt:
Suggested changes have been made.
I've also tested changes. Merged

PR #242, opened 2022-12-02
Add support for local.yml

andrewnicols:
This change adds support for a local yml file to include any customisations that users may wish to include as standard.

This may, for example include changes to existing containers, or it may include the addition of additional containers such as database admin tooling, or proxy servers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+local.yml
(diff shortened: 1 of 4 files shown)


scara:
Hi @andrewnicols,
LGTM!

Is there any reason not to kept aligned even [`moodle-docker-compose.cmd`](https://github.com/moodlehq/moodle-docker/blob/master/bin/moodle-docker-compose.cmd)?

TIA,
Matteo

andrewnicols:
Only because I totally forgot when I wrote this several months ago (and am still amazed that people use Windows for development).

I don't have a Windows machine to be able to test it I'm afraid.

stronk7:
Ah, and would be great if we can add the very same `ECHO "Including local options from...` to the Windows one. Sorry just saw it now.

andrewnicols:
> Only tiny suggestion that I'd do here is to add a comment about ensuring that the loading of the local.yml always remains the last one before calling compose. Just to avoid forgetting about that in the future.

Done :)

> Ah, and would be great if we can add the very same ECHO "Including local options from... to the Windows one. Sorry just saw it now.

And done - hope I got this right.

stronk7:
Thanks, looks perfect! Will test shortly...

stronk7:
Tested both under Unix and Windows, it worked great, so far!

PR #244, opened 2022-12-11
For sqlsrv, switch from 2017-latest to latest

stronk7:
Note that, right now, "latest" is exactly 2017-latest, so everything will continue working exactly the same.

But once we finish with the tests @ MDLSITE-6045, we'll make 2019-latest to become "latest".


stronk7:
Self-merging this. As said, note this doesn't change anything, because `latest` is still `2017-latest`. But this enabled us to make the jump to 2019 soon.

PR #245, opened 2022-12-17
Bump MySQL default to 8.0 and PostgreSQL to 13. And PHP defaults to 8.0

stronk7:
NOT TO BE MERGED YET!

They are the lowest versions supported for 4.2 and up.
    
Note that, at the time of writing this that's just a proposal, but that
way we can test if they work for all supported branches in advance.
    
So, here, we make MySQL 8.0 the default and create specific versions
for 5.6 and 5.7 in case anybody wants to use them.

Also, update GHA workflow to cover the new 401_STABLE branch and
update the master runs to use exclusively php80 (php81 will be added
once the branch is ready).

Finally, and not less important, change the default PHP version from 7.4 to 8.0. All
current supported versions support 8.0 (but 39_STABLE) and this can
be easily changed with `MOODLE_DOCKER_PHP_VERSION`. 

Note this is on hold until [MDL-74905](https://tracker.moodle.org/browse/MDL-74905) - Moodle 4.2 requirements is agreed.


stronk7:
As far as it seems that all the proposed changes for Moodle 4.2 are going to be accepted @ [MDL-74905](https://tracker.moodle.org/browse/MDL-74905)... applying them here a little bit in advance.

Ciao :-)

scara:
Hi @stronk7,
is there any reason to re-add `db.mysql.5.6.yml`?
Ref: https://github.com/moodlehq/moodle-docker/pull/225#issuecomment-1269014622

HTH,
Matteo

stronk7:
My only point was that the previous default `db.mysql.yml` supported both MySQL 5.6 and 5.7, and then the specific `db.mysql.8.0.yml` was, obviously, for MySQL 8.0

So, when making in this PR the 8.0 to become the default, we needed to continue supporting the 5.6 and 5.7 ones and that only can be done providing specific files for each (no matter they are 100% the same). Without them, they are sort of "unreachable" now.

Note that only Moodle 3.9.x (within the currently supported Moodle versions), supports MySQL 5.6. But until 3.9 falls out of support, it cannot hurt to keep it working, I think.

Ciao :-)

scara:
Hi @stronk7,
> Note that only Moodle 3.9.x (within the currently supported Moodle versions), supports MySQL 5.6. But until 3.9 falls out of support, it cannot hurt to keep it working, I think.

Thanks for taking the time to highlight the key point I was missing 💪 .

Ciao,
Matteo

PR #246, opened 2023-01-04
Ensure that we don't use bin logs by default

stronk7:
MySQL 8 comes with bin logs enabled, and that can end using big space on volumes for near nothing. So ensure that both MySQL and MariaDB, by default, disable them.

Note this can be changed whenever/if we want to test primary + replica configurations, but that's another story. For standalone runs better have bin logs disabled.

### DIFF
--- a/db.mariadb.yml
+++ b/db.mariadb.yml
@@ -11,6 +11,7 @@ services:
               --collation-server=utf8mb4_bin
               --innodb_file_per_table=On
               --wait-timeout=28800
+              --skip-log-bin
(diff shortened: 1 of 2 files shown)


scara:
Cross-ref: [MDL-74905 (comment)](https://tracker.moodle.org/browse/MDL-74905?focusedCommentId=970692&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-970692).

HTH,
Matteo

PR #254, opened 2023-03-17
Disable sqlsrv encryption by default

### LINKED ISSUE #253: Add support for new sqlsrv driver that is using odbc18
For more information, see https://github.com/moodlehq/moodle-php-apache/pull/164

Basically, odbc18 enables encrypted connections by default and our stack is not ready for that. So, what we'll be doing here, apart from other solutions is to mimic this from moodle-ci-runner:
https://github.com/moodlehq/moodle-ci-runner/commit/8dcbdc1ca751f48f09f0e2e594b563be2e29ee48

stronk7:
We don't need it, we don't want it. And, with
ODBC 18 it's enabled by default. Allowing/supporting it
in moodle-docker would require cert validations
and other stuff that we don't need for dev/test purposes.

Fixes #253

### DIFF
--- a/config.docker-template.php
+++ b/config.docker-template.php
@@ -13,6 +13,14 @@ $CFG->dbpass    = getenv('MOODLE_DOCKER_DBPASS');
 $CFG->prefix    = 'm_';
 $CFG->dboptions = ['dbcollation' => getenv('MOODLE_DOCKER_DBCOLLATION')];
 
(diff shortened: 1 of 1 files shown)


stronk7:
I'm self merging this because it doesn't harm and better have it ready for whenever we try an upgrade of the php images again.

PR #255, opened 2023-03-21
Switch to maintained multi-arch fork of mailhog/mailhog

stronk7:
Note that mailhog seems to be not much supported nowadays:
- https://github.com/mailhog/MailHog/issues/307
- https://github.com/mailhog/MailHog/issues/442

And there are a bunch of alternatives, also supporting multi-arch docker images and some more stuff:
- https://github.com/maildev/maildev
- https://github.com/axllent/mailpit

Surely we should move to one of them at some point.

### DIFF
--- a/service.mail.yml
+++ b/service.mail.yml
@@ -6,4 +6,4 @@ services:
     depends_on:
       - mailhog
   mailhog:
-    image: mailhog/mailhog
+    image: anatomicjc/mailhog


mattporritt:
Hi @stronk7 ,
I've tested this on an x86 and arm cpu machine. Multiarch works. I've also tested mailhog itself and it still works.
Merged.

Cheers,
Matt P

mattporritt:
Also I think we should use mailpit.
This is maintained, has multiarch containers and we've been using it internally for some testing to add better automation to behat tests.

PR #258, opened 2023-03-23
Replace mailhog with mailpit

### LINKED ISSUE #257: Switch from mailhog to mailpit
Mailhog is not really maintained anymore. We should switch to something else.
Mailpit is: https://github.com/axllent/mailpit
Plus it has multiarch docker images, and we're experimenting using it internally to help automate mail sending as part of behat.

mattporritt:
mailpit (https://github.com/axllent/mailpit) is a more maintained mail test client with a straightforward api.

Fixes #257

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ This repository contains Docker configuration aimed at Moodle developers and tes
 ## Features:
 * All supported database servers (PostgreSQL, MySQL, Micosoft SQL Server, Oracle XE)
 * Behat/Selenium configuration for Firefox and Chrome
-* Catch-all smtp server and web interface to messages using [MailHog](https://github.com/mailhog/MailHog/)
+* Catch-all smtp server and web interface to messages using [Mailpit](https://github.com/axllent/mailpit)
 * All PHP Extensions enabled configured for external services (e.g. solr, ldap)
 * All supported PHP versions
 * Zero-configuration approach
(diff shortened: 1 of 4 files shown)


mattporritt:
This replaces mailhog with mailpit. It is more or less a drop in replacement.

README.md
```diff
@@ -128,7 +128,7 @@ bin/moodle-docker-compose exec webserver php admin/cli/install_database.php --ag
 
 Notes:
 * Moodle is configured to listen on `http://localhost:8000/`.
-* Mailhog is listening on `http://localhost:8000/_/mail` to view emails which Moodle has sent out.
+* Mailpit is listening on `http://localhost:8025/_/mail` to view emails which Moodle has sent out.
```
stronk7, line 131:
```suggestion
* Mailpit is listening on `http://localhost:8000/_/mail` to view emails which Moodle has sent out.
```

I think `8000` is the correct public address that apache is proxying from internal (not published) `8025`.

stronk7 (commented):
Hi,

I think that with the tiny change suggested, this is pretty acceptable.

Nice one, thanks!

README.md
mattporritt: Ahh, thanks for the pickup. That is a leftover from an earlier version of the patch. I have fixed it now and force pushed

Cheers,
Matt P

andrewnicols (approved):

skodak:
this is a very annoying change, please do not change the contents of config.php unnecessarily, and if you do then please use more future-proof setting and docker image names such as:

`$CFG->smtphosts = 'mail:1025';`

PR #259, opened 2023-03-24
Add Matrix mock server

mattporritt:
In Moodle LMS 4.2 we will have support to connect to a Matrix instance. This patch adds support for the Matrix mock server we have developed. The mock server is required for unit and behat tests.

This mock has been implemented using the same stack (Symfony, Docker) as the existing BBB mock.


stronk7:
Wow, I thought that this had been merged long ago, in fact I went to use it when saw it had not!

My side it looks perfect (by comparing with the bbb-mock implementation, I don't know anything about the internals. So, with a couple of +1s (@mattporritt , @andrewnicols ), I'm happy proceeding with this.

Ciao :-)

stronk7:
Ok, going the brave way. Merging... thanks!

Issue #256, opened 2023-03-22
help: fresh install keeps logining out as time out


jpahullo:
Hi,

I reset my dev environment and I started from scratch using this repo, as before.

I build a Moodle 3.11 fresh install, building the set of containers, installing via CLI the database as suggested. All ok right now.

I log in as admin/test. Ok. I can click everywhere on the empty Moodle.

However, when I click on the Administration menu, it logs me out. And I cannot reenter into Moodle. No way.

I tried it for 3 times to start from scratch, and every time I got the same behaviour.

I can see that every failed attempt to log in, it creates a new session file, with the time stamp internal to the container apache webserver.

The loop for any admin login trial is:

1. "POST /login/index.php HTTP/1.1" 303 2141 "http://localhost/login/index.php"
2. "GET /login/index.php?testsession=2 HTTP/1.1" 303 1955 "http://localhost/login/index.php" 
  * admin is user.id = 2. so ok.
4. "GET /admin/search.php HTTP/1.1" 303 2025 "http://localhost/login/index.php"
  * enters into the admin_externalpage_setup()
  * then into the require_login()
  * finally in the lib/moodlelib.php::reguire_login() it detects that the user is not logged in. $USER->id = 0.
6. "GET /login/index.php HTTP/1.1" 303 1954 "http://localhost/login/index.php"
7. "GET /login/index.php HTTP/1.1" 200 8077 "http://localhost/login/index.php"
   * here the message is that my session was timed out. 

Incredibly. I cannot understand why I am facing this recurrent problem.

Any ideas? Thanks a lot in advance.

scara:
Hi @jpahullo,
could you share the exact sequence of commands used to run your Moodle 3.11 code, including your docker version and OS?
I'd like to replicate it in my own env.

HTH,
Matteo

jpahullo:
Thanks!

I'm at work. This afternoon I'll share the sequence.

Today, comming to work, I was thinking about the UTC timezone. I realized there were 1h of difference between host and container time (1 hour less in the container timezone). Maybe is this the reason of logging me out, since web browser says it is 1h after the session cookie? I'll try this this evening, mounting the local timezone file into the container, to unify timezones. 

I'll share all these details.

Thanks @scara for your quick reply!

Best,

Jordi

jpahullo:
Hi @scara,

Here are my findings.

Firstly I tried what I told you about the timezone. In my local.yml I have:

```yaml
version: "2"
services:
    webserver:
[...]

PR #260, opened 2023-03-24
 #256 - document how to prevent repeated log off

jpahullo:
After digging into the problem of #256, I have added a note over the manual testing section. 

For you, for me, for all of us.


stronk7:
Thanks @jpahullo !

PR #265, opened 2023-06-29
Split browser name before in case it uses tags

crazyserver:


stronk7:
Hi @crazyserver ,

change looks ok, good stuff!

Only question I've here is if, maybe, we also have to move the "split" code in `bin/moodle-docker-compose.cmd`. While the code managing the app for Windows is different from the Unix one... surely there we should be also looking for the `MOODLE_DOCKER_BROWSER_NAME`, and not for the whole `MOODLE_DOCKER_BROWSER` that can come with tags.

I think that, with that little detail covered... this will be perfect to be landed.

Ciao :-)

crazyserver:
Thanks Eloy! Can you review again? :-)

stronk7:
I've relaunched 2-3 jobs that failed (hopefully random ones). Code-wise, it seems that now everything is already in place, with the split happening before the browser name is used, so all ok.

stronk7:
Sold, thanks!

PR #275, opened 2023-11-13
Add some basic integration checks

### LINKED ISSUE #274: Cover mailpit with some integration tests
In #272 (thanks @paulholden !) it was detected that mailpit had stopped working. Something between v1.8.0 and current 1.10.0 has caused one env variable to be required in order to get it working.

It was fixed by #273 (thanks again, Paul!).

This is about to add some integration tests to GHA able to:
- Instantiate moodle-docker (with mailpit enabled).
- (surely it doesn't need to get a complete moodle/phpunit/behat site installed).
- Verify, with some manual (curl...) requests that mailpit (specifically css and js) is up and available.

And that's all, that will prevent future changes to cause it to stop working without notice.

Ciao :-)

stronk7:
The idea with these checks is to verify that the basic moodle-docker composition is able to get up and running and that the various entry points (php, web server, mailpit...) are available.

This can be expanded in the future to run more checks verifying other services or expectations.

Fixes #274

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,36 @@ name: moodle-docker CI
 on: [push, pull_request]
 
 jobs:
+  integration:
+    name: Integration checks
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checking out moodle-docker
+        uses: actions/checkout@v3
+
+      - name: Checking out moodle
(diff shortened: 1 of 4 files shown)


stronk7:
Self-merging this. It's passing ok and enables us to get those integration tests executed before everything else.

PR #276, opened 2023-11-23
Increase default PHP version to 8.1.

paulholden:
This is the minimum requirement for 4.4 as per MDL-78496.


stronk7:
LGTM, hanks! (waiting for tests to finish)

PR #282, opened 2024-03-12
#188 Fix scripts in MacOS

### LINKED ISSUE #188: [Mac] Unsupported grep flags used in `MOODLE_DOCKER_APP_RUNTIME` inference
The script implemented to guess the `MOODLE_DOCKER_APP_RUNTIME` env variable is using some flags that are not supported in Mac. It can be confusing because no error is thrown, instead the following message appears when running `moodle-docker-compose`:

```sh
$ bin/moodle-docker-compose exec webserver php admin/tool/behat/cli/init.php
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
  [-e pattern] [-f file] [--binary-files=value] [--color=when]
  [--context[=num]] [--directories=action] [--label] [--line-buffered]
  [--null] [pattern] [file ...]
```

The workaround for now is just to explicitly initialize the `MOODLE_DOCKER_APP_RUNTIME` variable in your local environment.

NoelDeMartin:
Fixes #188.

`grep -P` is not supported in MacOS, replacing it with `sed` should work everywhere.

### DIFF
--- a/bin/moodle-docker-compose
+++ b/bin/moodle-docker-compose
@@ -69,7 +69,7 @@ if [ -z "$MOODLE_DOCKER_APP_RUNTIME" ];
 then
     if [[ ! -z "$MOODLE_DOCKER_APP_PATH" ]];
     then
-        appversion="$(cat $MOODLE_DOCKER_APP_PATH/package.json | grep -oP '"version": "\K\d\.\d\.\d{1,2}(-\w+)?(?=")')"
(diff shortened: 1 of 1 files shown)


stronk7:
Merged, thanks!

NoelDeMartin:
Thanks for merging @stronk7. We have tested these in MacOS and they work without installing any special packages, so hopefully that makes it work everywhere.

PR #290, opened 2024-04-22
Add MOODLE_404_STABLE GHA tests

stronk7:
Basically, duplicate all 403 entries to 404 and amend highest/lowest PHP versions for all the new ones.

But for the app tests, where we (agreed with the apps team)
just replace the previous (403) with 404 in the first 2
groups of tests, and add in the 3rd group.


stronk7:
It seems that all the new jobs passed, ok, so self-merging.

Ciao :-)

PR #291, opened 2024-05-06
Gitpoddata

sarjona:
Added a few improvements to the Gitpod integration:

- Set gitpod folder permission to install plugins
- Initialise gitpod with data when DATAFILE variable is given
- Add adminer to gitpod (only when INSTALLADMINER variable is given)
- Add parameter to speed gitpod clone. By default, a shallow clone is used when downloading the moodle.git repository. For a full clone, the parameter CLONEALL can be used.

PR #292, opened 2024-05-10
Gitpod: Improve git performance

sarjona:
In https://github.com/moodlehq/moodle-docker/pull/291, the CLONEALL parameter was added, in order to improve the performance. However, I forgot to also update the piece of code where the branch is updated to the latest (to remove any change done) every time a workspace is open again. 
A tiny improvement can be added here to reduce the time :-)


stronk7:
Cannot but trust you :-)

PR #293, opened 2024-07-01
Fall back to ionic7 app runtime

NoelDeMartin:
We've released 4.4.0 version of the app, so `ionic7` can now be used as the default runtime.


stronk7:
So, it seems that we need to do something, in GHA tests,  with old branches? (311 and 400) to keep them working with ionic 5? Or just remove them?

NoelDeMartin:
@stronk7 It should work now, I decided to keep them running against old versions to make sure that the ionic5 runtime still works in case someone needs it. But eventually we could remove it as well because none of the current versions of the app (stable, development) use it anymore.

stronk7:
Thanks, @NoelDeMartin,

looks ok. I'll wait for all tests to finish (they are looking perfect) and will merge this then (1h aprox).

Ciao :-)

stronk7:
Ok, there was a PHPUnit failure (random) and I've relaunched the job. No matter of that, I'm going to merge this now, so it's available for the masses.

Ciao :-)

Issue #294, opened 2024-07-01
Allow Behat timeout be increased via an environment variable


NeillM:
It is useful to be able to set the behat timeout factor, especially if running docker on a slower machine)

We could allow `$CFG->behat_increasetimeout` to be set by environment variable

stronk7:
Closing this because #295 has been merged, thanks!

PR #295, opened 2024-07-01
Allow the Behat timeout factor to be configured

NeillM:
Changes for #294 we have been running a version of this patch on our automatic testing server for years now.


NeillM:
I'm guessing the app test failures here are related to the reason [Fall back to ionic7 app runtime](https://github.com/moodlehq/moodle-docker/pull/293) was created

NoelDeMartin:
@NeillM The fix for the app failures has been merged, if you rebase you should get rid of the app errors.

stronk7 (approved):
Thanks @NeillM , looks spot-on!

Was going to comment about when 0 (zero) or 1 is the best default, but both lead to the same, from memory (aka, original timeout applied), so no problem, so far.

PR #298, opened 2024-08-23
Bump Selenium to version 4

andrewnicols:


skodak:
this makes sense, the recent 4.2x.x releases seem pretty solid and I did not encounter any problems with them yet

junpataleta (approved):
Thanks, Andrew. Looks good to me

PR #306, opened 2024-11-18
MECONTRACK-242 - Improvement to k1many.sh

nicolas-dalpe:
MECONTRACT-242 - Set unique session cookie name to avoid login issue.
MECONTRACT-242 - DRY COMPOSE_PROJECT_NAME.
MECONTRACT-242 - Added error_message() decorator function.
MECONTRACT-242 - Added info_message() decorator.
MECONTRACT-242 - Rename the Moodle instances according to folder's name.
MECONTRACT-242 - Use the folder name as the project name so it is easier to recognize in the Docker UI.
MECONTRACT-242 - Ensure the arguments passed in are checked properly; Ensure exit commands are used rather than returns; Group commands to reduce redundancy;


nicolas-dalpe:
To be deleted. Sorry about that.

PR #310, opened 2024-12-02
Increase Moodle 5.0 requirements for PHPUnit too

sarjona:
This commit adds missing values in the phpunit matrix (missed in the last commit and raised by the GHA).

5.0 agreed environment requirements:

    - PHP 8.2
    - PostgreSQL 14
    - MySQL 8.4
    - MariaDB 10.11.0

Apart from this, the Moodle 4.5 branch has also been added to GHA.


sarjona:
Hi @paulholden!
Whenever you can, it would be great if you could also peer-review this (to increase requirements for the PHPUnit jobs too).
The app failure can be ignored because it's unrelated to the patch and Dani Palou from the apps team is looking at it ;-)

paulholden (approved):

PR #312, opened 2024-12-10
Allow Moodle app port to be configured

NeillM:
Before this change the moodle-app was always put onto port 8100, this makes it impossible (without using local.yml) to have multiple instances of moodle-docker running if they also have a copy of the app running.

After this change while be default nothing should change you may now select the exposed port, or disable it altogether using MOODLE_DOCKER_APP_PORT

Before this change MOODLE_DOCKER_APP_PORT was used internally to select the source port for the app, this is now done using MOODLE_DOCKER_APP_INTERNAL_PORT because that name fits better with the other environment variables that let the user set the port available on the host.


NeillM:
One of the unit tests failed... since I did not change any Moodle code I'm not sure that is related to this change.

NeillM:
I have rebased the change and fixed the conflict.

PR #316, opened 2025-02-21
Support Python Machine learning backend

HuongNV13:
The patch adds support for Python Machine learning backend so we can run PHPUnit for `core_analytics_testsuite`


lameze:
Hi @HuongNV13 thanks for adding the support for Python machine learning backend.

Overall everything looks good, I just have a few questions: 
1. Do we also need to add the support for Behat? If yes, maybe we should rename this $MOODLE_DOCKER_PHPUNIT_MLBACKEND to something less specific to PHPUnit?
2. If this is specific to PHPUnit should it work with this [MOODLE_DOCKER_PHPUNIT_EXTRAS](https://github.com/moodlehq/moodle-docker/pull/316/commits/b2b02fcc230066f49723dd5b0cfb957c10c3a7f6#diff-050e770d2e7b0164ab02cf3ecd528490a0bc208f9feb564e0ac2fe919c13a2faR99) and be placed in that if block alongside all other extras (solr, redis, ldap and etc)?
3. Do you think it worth adding that new env. variable to the list of [Environment variables](https://github.com/moodlehq/moodle-docker?tab=readme-ov-file#environment-variables)?
4. I've had a look on the failed builds related to this PR and they seem unrelated, could you please double check?

Anyway, great job legend.

Cheers

HuongNV13:
Thanks, @lameze for taking a look at this one.
Yes, this one is for both Behat and PHPUnit, so I have renamed it to `MOODLE_DOCKER_MLBACKEND`
I also added a new environment variable to the readme file.

The failures are not related to this patch, we might need to update the MSSQL YML file to make the tests back to green.

Thanks,

lameze:
Hi @HuongNV13 thanks for making those changes.

Given this can be used by Behat and PHPUnit, then it makes sense to have a [separated section](https://github.com/moodlehq/moodle-docker/pull/316/commits/f6c1b50bcf8fd910d6feddec443d2918287e00e6#diff-050e770d2e7b0164ab02cf3ecd528490a0bc208f9feb564e0ac2fe919c13a2fa) in config.php instead of adding it to the [phpunit-external-services.yml](https://github.com/moodlehq/moodle-docker/blob/3a4e5905f7082dbee1254fa106caa552757fe323/phpunit-external-services.yml).

Happy to get this merged now.

Cheers,

Simey

PR #321, opened 2025-03-31
Disable Apache FallbackResource for faildumps

### LINKED ISSUE #320: Problems accessing faildump on 5.0dev - looks like routing conflict?
I'm not quite sure what is happening but I suspect that routing in 5.0dev is trying to intercept a request to http://localhost:6002/_/faildumps/ because I get errors about capabilities and contexts and I am not even trying to access Moodle

(I have `MOODLE_DOCKER_WEB_PORT=6002`)

Image

Image

andrewnicols:
Per docs in https://httpd.apache.org/docs/trunk/mod/mod_dir.html#fallbackresource the FallbackResource can be disabled using the {{disabled}} argument where it is set in a parent path.

Fixed #320

### DIFF
--- a/assets/web/apache2_faildumps.conf
+++ b/assets/web/apache2_faildumps.conf
@@ -2,5 +2,8 @@
     Options +Indexes
 </Directory>
 
-Alias /_/faildumps /var/www/behatfaildumps
+<Location "/_/">
+    FallbackResource disabled
+</Location>
 
+Alias /_/faildumps /var/www/behatfaildumps
--- a/assets/web/apache2_faildumps.conf
+++ b/assets/web/apache2_faildumps.conf
@@ -7,3 +7,15 @@
 </Location>
 
 Alias /_/faildumps /var/www/behatfaildumps
+
+# Set the handler to use for files whose extension ends in `.php`.
+# This is a workaround for https://github.com/docker-library/php/issues/1576
+# The FallbackResource is a default Handler action which is only used if a Handler has not already been applied.
+# The docker/php image sets the handler unconditionally for any file whose name matches `.php`.
+# This configuration block unsets the default handler, and instead only applies it if the file exists.
+<FilesMatch \.php$>
+  SetHandler none
+  <If "-f %{REQUEST_FILENAME}">
+    SetHandler application/x-httpd-php
+  </If>
+</FilesMatch>
--- a/base.yml
+++ b/base.yml
@@ -5,7 +5,7 @@ services:
       - db
     volumes:
       - "${MOODLE_DOCKER_WWWROOT}:/var/www/html"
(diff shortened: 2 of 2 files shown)


andrewnicols:
Pinging @marinaglancy,

This should solve the issue. The docker image places the `FallbackResource` configuration in a file named `fallback.conf` which was being loaded _after_ the `apache-faildump.conf` file and therefore overriding it, despite being less specific.

I've renamed the file to `moodle-faildump.conf` which should solve the issue. Can you test again please?

marinaglancy:
it works in the way that the `/_/faildumps/` listing is available, but does not it remove the fallback to r.php completely?

andrewnicols:
Yes. That is the intended behaviour

Andrew Lyons
Principal Architect

Moodle HQ - moodle.com [image: Moodle - Empowering educators to improve our
world]

andrewnicols:
Okay, replying now to you redited comment instead of the mail I recived

So it's working in that the fallback is not applied to the `_` Location. I'll look into the .php side of things.

andrewnicols:
le sigh: https://bz.apache.org/bugzilla/show_bug.cgi?id=52403#c7

> If you use SetHandler or AddHandler with FallbackResource non-existent PHP file do not use the FallbackResource. With AddType instead of SetHandler or AddHandler the FallbackResource works as expected.

Response:
> This would be expected. The alternate handler is responsible for the content. The FallbackResource applies only to the core file system (default) handler.

Totally undocumented behaviour.

There seem to be a couple of options to address this:

1. Switch from `SetHandler` to `AddType`
2. Wrap the `SetHandler` call in an conditional
3. Switch to `mod_rewrite` (which is not recommended for this usage)
4. Switch to `ErrorHandler` (which is not recommended for this usage)

andrewnicols:
I've raised https://github.com/docker-library/php/issues/1576 against the upstream Docker image. I've also raised a documentation issue against Apache: https://bz.apache.org/bugzilla/show_bug.cgi?id=69640

marinaglancy:
Sorry for editing the comment. I thought the fallback was not working but I was checking the "wrong" url. Good job finding the cause of it.

andrewnicols:
If you're happy with the solution I've proposed here, please review + merge :0)

marinaglancy:
I do not have formal permissions to review or merge in this repo. 

I confirm that both patches fix respective issues.

paulholden:
Thanks both!

andrewnicols:
Yes, but overrides need to be location specific. I’m hoping the upstream
images updates to use the IfMarch config too which will negate the second
issue.

Andrew Lyons
Principal Architect

Moodle HQ - moodle.com [image: Moodle - Empowering educators to improve our
world]

PR #326, opened 2025-05-20
Make the usage example branch agnostic

andrewnicols:


fulldecent:
This change makes it harder for people to get started.

Rather than copy and pasting, it requires each person to copy-paste and think.

---

It would be much more scalable if Moodle, Inc. would think and update this once or twice per year. Then the millions of people that use this project could copy and paste without thinking.

---

And the much much more better solution is that Moodle, Inc [maintains a MOODLE_STABLE branch](https://tracker.moodle.org/browse/MDLSITE-8072). This requires only the release manager who publishes new Moodle stable versions to think, and only once per release (which should ideally be automated). And then neither the maintainers here, the maintainers at any project, nor the users of any project will need to think——all they need to do is copy paste.

fulldecent:
An important function of DEVREL is to make projects easier, and give them examples that work with good defaults that can be copy-pasted.

PR #330, opened 2025-07-29
Check nvmrc node version names

crazyserver:
Moodle App now uses version name on .nvmrc file to determine the node version to use.

To select the node image the slash (/) should be replaced by a hyphen (-)


crazyserver:
Test failing are unrelated to the issue.

sarjona (approved):
Hola @crazyserver!
Thanks for working on this. The patch for Gitpod also looks good and works as expected so I'm happy to approve and merge it :-)

PR #342, opened 2025-12-04
Add support for docker compose v5. See https://docs.docker.com/com…

tmas0:
Hi,

This pull request contains the minimum and necessary changes to add support for Docker Compose v5. See the releases notes: https://docs.docker.com/compose/releases/release-notes/

Bests


NeillM:
If we make the regular expression:

`^v?([2-9]|1[0-9])`

I think it should match any version apart from version 1; i.e. so it will not fail if they go to version 6, or version 10 for example.

[See it in action](https://regexr.com/8irei)

